PR #405, opened 2022-06-15
Feat/corestaxzaps

princetonbishop:
# Description
What does this PR solve?
- Zap contracts for temple, core and stax

Fixes # (issue)

# Checklist


scooprinder (commented):
Just a couple of small nits for now. Will continue to look at the zap into LP funcs but I think its in a good enough state to get tests rolling. 

If you wanted to use foundry, I believe I finally solved the uniswap/lib issue so it should now play nicely with our hardhat set up, if you wanted to go down the foundry route.

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+  /**
+   * @notice recover token or ETH
+   * @param _token token to recover
+   * @param _to receiver of recovered token
+   * @param _amount amount to recover
+   */
+  function recoverToken(address _token, address _to, uint256 _amount) external onlyOwner {
+    require(_to != address(0), "Invalid receiver");
+    if (_token == address(0)) {
+      // this is effectively how OpenZeppelin transfers eth
```
frontier159, line 150: Any reason why we aren't using SafeERC20?

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+      fromAmount,
+      toToken,
+      amountOutMin,
+      swapTarget,
+      swapData
+    );
+  }
+
+  function zapLiquidityBalancerPoolFor(
+    address _fromToken,
```
frontier159, line 197: Wonder if it's worth breaking up the balancer | curve | uni into separate implementation files? Is this starting to hit up against the bytesize limits or ok?

protocol/contracts/TempleZaps.sol
```diff
@@ -0,0 +1,190 @@
...
+    uint256 fromAmount,
+    address toToken,
+    uint256 amountOutMin,
+    address swapTarget,
+    bytes calldata swapData
+  ) external returns (uint256 amountOut);
+}
+
+
+contract TempleZaps is Ownable {
```
frontier159, line 78: Is this strictly required, or really just a helper so there's only one approval step?

Would the idea be to implement an equivalent for Stax?

frontier159:
Mostly just smaller nits - looks great, but i'm keen to see tests to understand the actual data that flows through...writing tests won't be a small task!

protocol/contracts/GenericZap.sol
scooprinder: You can run `yarn size` and it'll print out the size of all the contracts - was super useful when splitting up OpsManager in the core contracts.

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+            reserveIn * ((userIn * 3988000) + (reserveIn * 3988009))
+        ) - (reserveIn * 1997)) / 1994;
+  }
+
+  function _getSwapAmount(uint256 amountA, uint256 reserveA) internal pure returns (uint256) {
+    return (sqrt(amountA * ((reserveA * 3988000) + (amountA * 3988009))) - (amountA * 1997)) / 1994;
+  }
+
+  // babylonian method (https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Babylonian_method)
+  function sqrt(uint y) internal pure returns (uint z) {
```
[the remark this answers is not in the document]
princetonbishop: this is from uniswap lib. will add attribution

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+    address _toToken,
+    address _swapTarget,
+    bytes memory _swapData
+  ) internal returns (uint256 amountBought) {
+    if (_swapTarget == WETH) {
+      _depositEth(_fromAmount);
+      return _fromAmount;
+    }
+
+    uint256 valueToSend;
```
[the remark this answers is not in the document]
princetonbishop: `uint256 valueToSend` is cheaper than `uint256 valueToSend = 0`

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+      _poolId,
+      _recipient,
+      _swapTarget,
+      _swapData,
+      _zapLiqRequest,
+      _request
+    );
+  }
+
+  function zapLiquidityBalancerPool(
```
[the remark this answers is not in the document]
princetonbishop: not quite sure i understand what you mean in the first part. there's already a `zapLiquidityBalancerPoolFor` function (including the other lp pools).

I will add a `zapInFor`

protocol/contracts/GenericZap.sol
princetonbishop: `yarn size` shows `24.052` . very close. will do some splitting

protocol/contracts/GenericZap.sol
```diff
@@ -0,0 +1,918 @@
...
+  ) internal {
+    require(approvedTargets[_fromToken][_swapTarget] == true, "Unsupported token/target");
+
+    // pull tokens
+    _pullTokens(_fromToken, _fromAmount);
+
+    uint256 nCoins = curveFactory.get_n_coins(_pool);
+    address[] memory coins = new address[](nCoins);
+    uint256 fromTokenIndex = nCoins; // set wrong index as initial
+    uint256 otherTokenIndex;
```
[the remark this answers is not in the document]
princetonbishop: yep, good call

protocol/contracts/TempleZaps.sol
princetonbishop: temple has specific requirements to zapping into vaults, hence separation of concerns with a dedicated contract which utilizes functions from generic zaps.
yes, stax would have its own after we specify requirements and there are unique ones. so far, stax can use everything in the generic zaps

protocol/contracts/GenericZap.sol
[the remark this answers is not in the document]
frontier159: I guess typically in the non-contract space I'd write as:
```
  function zapLiquidityBalancerPoolFor(...) public payable whenNotPaused {
    require(approvedTargets[_fromToken][_swapTarget] == true, "Unsupported token/target");

    _pullTokens(_fromToken, _fromAmount);

    _zapLiquidityBalancerPool(
      _fromToken,
      _fromAmount,
      _poolId,
      _recipient,
      _swapTarget,
      _swapData,
      _zapLiqRequest,
      _request
    );
  }

  function zapLiquidityBalancerPool(...) external payable {
    zapLiquidityBalancerPoolFor(..., msg.sender, ...);
  }
```

So the requires, modifiers, pullTokens etc are guaranteed to be consistent / non-duplicated logic too. I understand there's a tiny bit more gas because another function redirect, but would expect that to be very small.

frontier159 (commented):
Sorry - I didn't get a chance to get much of the way through this. I'll pick up again tomorrow and will let you know when complete.

Also - this looks empty?
    protocol/contracts/zaps/libs/Transfer.sol

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+
+  // fromToken => swapTarget (per curve, univ2 and balancer) approval status
+  mapping(address => mapping(address => bool)) public approvedTargets;
+
+  receive() external payable {
+    require(msg.sender != tx.origin, "Do not send ETH directly");
+  }
+
+  /**
+    @dev Adds or removes an approved swapTarget
```
frontier159, line 23: nit: These should be @notice, not @dev?

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+    for (uint256 i = 0; i < _length; i++) {
+      approvedTargets[_tokens[i]][_targets[i]] = _isApproved[i];
+    }
+  }
+
+  /**
+    @dev Toggles the contract's active state
+     */
+  function toggleContractActive() external onlyOwner {
+    paused = !paused;
```
frontier159, line 46: add an event?

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+    */
+  function setApprovedTargets(
+    address[] calldata _tokens,
+    address[] calldata _targets,
+    bool[] calldata _isApproved
+  ) external onlyOwner {
+    uint256 _length = _isApproved.length;
+    require(_targets.length == _length && _tokens.length == _length, "Invalid Input length");
+
+    for (uint256 i = 0; i < _length; i++) {
```
frontier159, line 37: If the existing `approvedTargets` (before this is called) is larger then the new length, then there'll be some stale/residual approved targets I think?

Or is the expectation that the caller sets to these to `_isApproved=false` for the remaining? Bit of a gotcha

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+    paused = !paused;
+  }
+
+  /**
+    @notice Approve a token for spending with finite allowance
+    @param _token The ERC20 token to approve
+    @param _spender The spender of the token
+    @param _amount The allowance to grant to the spender
+     */
+  function _approveToken(
```
frontier159, line 55: Is this required as opposed to adding in-line where required? extra gas for a pretty small bit of syntactic sugar?

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+    address _token,
+    address _spender,
+    uint256 _amount
+  ) internal {
+    IERC20(_token).safeIncreaseAllowance(_spender, _amount);
+  }
+
+  function _transferToken(IERC20 _token, address _to, uint256 _amount) internal {
+    uint256 balance = _token.balanceOf(address(this));
```
frontier159, line 64: This check isn't strictly required, since the underlying `safeTransfer()` will fail too. But I guess this gives a simple/consistent message rather than being dependent on the underlying token implementation/revert message?

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,103 @@
...
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+abstract contract ZapBase is Ownable {
+  using SafeERC20 for IERC20;
+
+  bool public paused;
+
+  uint256 internal constant DEADLINE = 0xf000000000000000000000000000000000000000000000000000000000000000;
```
frontier159, line 13: I don't know enough about the AMM - but is it ok that this has such a large deadline?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,697 @@
+pragma solidity ^0.8.4;
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import "./ZapBase.sol";
+import "./libs/Swap.sol";
+//import "./libs/Executable.sol";
```
frontier159, line 6: cleanup

protocol/test/temple-zaps.ts
```diff
@@ -0,0 +1,2048 @@
...
+    });
+
+    it("recovers token", async () => {
+      // transfer frax to zaps contract
+      const frax = IERC20__factory.connect(FRAX, fraxSigner);
+      await frax.transfer(genericZaps.address, 1000);
+            
+      // recover
+      const checkSumedAddr = ethers.utils.getAddress(FRAX);
+      await expect(genericZaps.recoverToken(FRAX,  ownerAddress, 1000))
```
frontier159, line 151: Need to test recovering ETH?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,697 @@
...
+   * @param _token token to recover
+   * @param _to receiver of recovered token
+   * @param _amount amount to recover
+   */
+  function recoverToken(address _token, address _to, uint256 _amount) external onlyOwner {
+    require(_to != address(0), "Invalid receiver");
+    if (_token == address(0)) {
+      // this is effectively how OpenZeppelin transfers eth
+      require(address(this).balance >= _amount, "Address: insufficient balance");
+      (bool success,) = _to.call{value: _amount}("");
```
frontier159, line 58: Hmm we can't recover ETH from stax contracts...worth us considering adding this to the Transfer lib?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,697 @@
...
+  ) public payable whenNotPaused {
+    require(approvedTargets[_fromToken][_swapTarget] == true, "Unsupported token/target");
+
+    _pullTokens(_fromToken, _fromAmount);
+
+    uint256 tokenBoughtIndex;
+    uint256 amountBought;
+    (address[] memory poolTokens,,) = balancerVault.getPoolTokens(_poolId);
+    bool fromTokenIsPoolAsset = false;
+    uint i = 0;
```
frontier159, line 250: This can be moved into the `for(uint i=0; ...)` now?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,697 @@
...
+    address _pool,
+    address _recipient,
+    uint256 _minLiquidityOut,
+    bool _useAltFunction,
+    uint256[] memory _amounts
+  ) internal returns (uint256 liquidity) {
+    bool success;
+    bytes memory data;
+    if (_useAltFunction) {
+      //liquidity = ICurvePool(_pool).add_liquidity(_amounts, _minLiquidityOut, false, _recipient);
```
frontier159, line 497: Specifying the direct hex to save the `bytes4(keccak256("add_liquidity.....))` call? What's the extra bool param?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,697 @@
...
+  function _addLiquidityUniV2(
+    address _pair,
+    address _recipient,
+    uint256 _amountA,
+    uint256 _amountB,
+    uint256 _amountAMin,
+    uint256 _amountBMin,
+    bool _shouldTransferResidual
+  ) internal {
+    address tokenA = IUniswapV2Pair(_pair).token0();
```
frontier159, line 520: I don't fully understand the stack-too-deep, but why not add tokenA/tokenB down too?

protocol/contracts/zaps/GenericZap.sol
princetonbishop: 1/ yes.
2/ the extra bool param is a special case for some base 2 plain pools eg. fxsCvxFxs pool
I've added this explanation to documentation. will push commit soon

protocol/contracts/zaps/GenericZap.sol
princetonbishop: yes, it's worth adding that

protocol/test/temple-zaps.ts
princetonbishop: sure, will add

protocol/contracts/zaps/ZapBase.sol
princetonbishop: modifier in AMM checks for `deadline >= blocktimestamp`

protocol/contracts/zaps/ZapBase.sol
princetonbishop: yes, it will bubble the error `SafeERC20: low-level call failed"`. above helps us to fail faster and save some gas.
Although you can also argue that assuming the tx is successful, the extra check cost some gas.

protocol/contracts/zaps/ZapBase.sol
princetonbishop: yh good call. there were some cases i had to use inline for stack management.

protocol/contracts/zaps/ZapBase.sol
princetonbishop: not sure if I understand you completely. but since it's a mapping, default for a `Map[A][B]` pair is `false`. so the new incoming approvals can be of any length and map values will either be updated/reset or new entries set.

protocol/contracts/zaps/libs/Executable.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity ^0.8.4;
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+/// @notice An inlined library function to add a generic execute() function to contracts.
+/// @dev As this is a powerful function, care and consideration needs to be taken when 
+///      adding into contracts, and on who can call.
+library Executable {
+    //error UnknownFailure();
```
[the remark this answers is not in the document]
princetonbishop: yes, I plan to add a new issue to upgrade entire temple codebase to use custom errors

.github/workflows/protocol-tests.yaml
```diff
@@ -19,6 +19,9 @@ jobs:
         working-directory: protocol
       - run: yarn test
         working-directory: protocol
+        env:
+          TESTS_FORK_BLOCK_NUMBER: ${{ secrets.TESTS_FORK_BLOCK_NUMBER }}
```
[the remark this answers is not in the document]
princetonbishop: true. the tests don't use this value anymore. will remove

protocol/contracts/zaps/GenericZap.sol
princetonbishop: it's because of the way variables are cycled. by scoping the rest of code, those variables are handled differently pertaining to the stack iirc.

protocol/contracts/zaps/ZapBase.sol
frontier159: At t0:
```
setApprovedTargets(
    [FRAX, TEMPLE, UNI, FXS],
    [ZEROEX, TEMPLE_AMM, TEMPLE_AMM, ZEROEX],
    [true, true, true, true]
);
```

At t1:
```
setApprovedTargets(
    [cvxFxs, TEMPLE],
    [ZEROEX, TEMPLE_AMM],
    [true, true]
);
```

I think in this case the UNI, FXS are still left as approved targets?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,694 @@
...
+contract GenericZap is ZapBase {
+
+  IUniswapV2Router public immutable uniswapV2Router;
+  ICurveFactory private immutable curveFactory = ICurveFactory(0xB9fC157394Af804a3578134A6585C0dc9cc990d4);
+  IBalancerVault private immutable balancerVault = IBalancerVault(0xBA12222222228d8Ba445958a75a0704d566BF2C8);
+
+  struct ZapLiquidityRequest {
+    uint248 firstSwapMinAmountOut;
+    bool useAltFunction;
+    uint248 poolSwapMinAmountOut;
```
scooprinder, line 22: uint248 for struct packing?

protocol/contracts/zaps/ZapBase.sol
scooprinder: I'd argue that is how we'd want a setter to work? If I don't explicitly pass in a param, I'd be surprised if the setter modified that param?

I'm a real stickler for no side effects in setters; can change the name if that side effect is expected though

protocol/contracts/zaps/ZapBase.sol
```diff
@@ -0,0 +1,104 @@
...
+  address private constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+  uint256 internal constant DEADLINE = 0xf000000000000000000000000000000000000000000000000000000000000000;
+
+  // fromToken => swapTarget (per curve, univ2 and balancer) approval status
+  mapping(address => mapping(address => bool)) public approvedTargets;
+
+  event SetContractState(bool paused);
+
+  receive() external payable {
+    require(msg.sender != tx.origin, "Do not send ETH directly");
```
scooprinder, line 22: nit: prefix error messages with contract name (can use ZB or similar abbreviation if we're attempting to stick < 32 chars)

protocol/contracts/zaps/TempleZaps.sol
```diff
@@ -0,0 +1,504 @@
...
+import "./interfaces/ITempleStableRouter.sol";
+import "./interfaces/IGenericZaps.sol";
+import "./interfaces/IVaultProxy.sol";
+import "./interfaces/IVault.sol";
+
+
+contract TempleZaps is ZapBase {
+  using SafeERC20 for IERC20;
+
+  address private constant FRAX = 0x853d955aCEf822Db058eb8505911ED77F175b99e;
```
scooprinder, line 19: is this used? Should be immuatable and passed via ctor to help with testnet

protocol/contracts/zaps/TempleZaps.sol
```diff
@@ -0,0 +1,504 @@
...
+   * @param _vault Target core vault
+   * @param _fromToken The token used for entry (address(0) if ether)
+   * @param _fromAmount The amount of fromToken to zap
+   * @param _minTempleReceived Minimum tokens out after first DEX swap
+   * @param _stableToken Supported temple pair stable token
+   * @param _minStableReceived Minimum stable token to receive
+   * @param _swapTarget Execution target for the swap
+   * @param _swapData DEX data
+   */
+  function zapTempleFaithInVault(
```
scooprinder, line 190: We're no longer supporting faith in core (as in, not allowing any more deposits via faith) so we can remove any faith related code from this contract. (sorry man)

protocol/contracts/zaps/libs/EthConstants.sol
```diff
@@ -0,0 +1,6 @@
+pragma solidity ^0.8.4;
+// SPDX-License-Identifier: GPL-3.0-or-later
```
scooprinder, line 2: nit: license for code we've authored should be AGPL, if possible

protocol/test/temple-zaps.ts
```diff
@@ -0,0 +1,2056 @@
...
+let fraxSigner: Signer;
+let ownerAddress: string;
+let aliceAddress: string;
+let alanAddress: string;
+let balancerVault: IBalancerVault;
+let poolId: string;
+let bptPoolToken: string;
+
+
+describe.only("Temple Stax Core Zaps", async () => {
```
scooprinder, line 61: nit: remove only before merge

protocol/test/temple-zaps.ts
```diff
@@ -0,0 +1,2056 @@
...
+  before(async () => {
+    // as real-time live data is being queried from 0x API, there's the possibility that prices may differ from forked mainnet tests (due to block number)
+    // and hence tokens out (in tests) may differ from real expected values 
+    // use most recent block number in tests (with enough block confirmations)
+    await resetFork(await getLatestBlockNumberWithEnoughConfirmations());
```
scooprinder, line 66: have we tried running the entire test suite with this? My concern is any tests following this suite will run with forked chain (not necessarily the worst thing but is a performance hit)

scooprinder (commented):
Just a few small nits and questions. Looking really good @princetonbishop !

protocol/contracts/zaps/ZapBase.sol
frontier159: Perhaps just a naming thing to me then - when I first read 'setApprovedTargets()', and that it takes an array, I think it implies that's the full set of targets - anything else is disabled

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -0,0 +1,694 @@
...
+        break;
+      }
+      unchecked { i++; }
+    }
+    require(tokenBoughtIndex != nCoins, "Invalid swap");
+
+    return (tokenBoughtIndex, bal - balancesBefore[tokenBoughtIndex]);
+  }
+
+  // function _depositEth(
```
frontier159, line 641: Not sure if you were in the process of cleaning up?

protocol/test/temple-zaps.ts
```diff
@@ -0,0 +1,2056 @@
...
+  console.log(
+    `Starting Temple: ${ethers.utils.formatUnits(balanceBefore, 18)}`
+  );
+  console.log(`Selling ${fromAmount} ${symbol}`);
+
+  // Approve token
+  if (fromTokenAddress !== ETH) {
+    await tokenContract.approve(
+      templeZaps.address,
+      ethers.utils.parseUnits('1000111', decimals)
```
frontier159, line 1523: Meant to remain as this odd amount? Are you intending the `console.log()` to remain?

protocol/test/temple-zaps.ts
```diff
@@ -0,0 +1,2056 @@
...
+    const response = await axios.get(url);
+    ({
+      data: { data: swapCallData, price, guaranteedPrice, gas, estimatedGas },
+    } = response);
+
+    console.log(`Price of ${symbol} in FRAX: ${price}`);
+    console.log(`Guaranteed price: ${guaranteedPrice}`);
+  }
+
+  // todo: swap on temple zaps. probably update templezaps to use stable router after first swap
```
frontier159, line 1683: A few todo's and comments to check

protocol/contracts/zaps/GenericZap.sol
princetonbishop: yes ser

protocol/contracts/zaps/ZapBase.sol
princetonbishop: I'm with scoop's arguments. if I don't pass in a token/target pair, the old value should remain as is in storage

protocol/contracts/zaps/TempleZaps.sol
princetonbishop: not used, will remove.
@scooprinder re testnet, given the nature of real-time dex data (from 0x) for the contracts and liquidity pools to function as expected, do you still think testnet should be on the table? thoughts?

protocol/test/temple-zaps.ts
princetonbishop: well before, I had run the entire test suite even with the forked state and remaining tests afterwards finished successfully.
But will double-check and worst case scenario i will rename this test to be last or use a global `after()` hook

protocol/contracts/zaps/GenericZap.sol
princetonbishop: missed that, was coding in a zombie state last night

protocol/test/temple-zaps.ts
princetonbishop: the initial intention was to leave the console logs only for the price info.
@scooprinder you also got any views on this?

protocol/contracts/zaps/TempleZaps.sol
frontier159: It's going to be really hard to maintain using testnet for STAX - eg for GMX they don't really have a maintained testnet that I can see.

princetonbishop: yes I agree with you @frontier159 , testing real-time data on testets is a hustle. hence why i ignored testnets until now.

protocol/test/temple-zaps.ts
princetonbishop: double-checked and all tests pass locally. also, included an `after()` hook

frontier159 (approved):
LGTM - great work!

scooprinder (approved):
LGTM - awesome job, glad to see this one get over the line @princetonbishop

PR #433, opened 2022-07-08
Delete/Deprecate unused contracts

butlerji:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist
- [x] Code follows the style guide
- [x] I have performed a self-review of my own code
- [x] New and existing tests pass locally
- [x] This PR is targeting the correct branch

Issue #435, opened 2022-07-08
Single, Unified List of Deployed and actively used contracts


butlerji:
Currently there is a list in scripts/deploys/helpers.ts - and a semantic used contracts by dapp list (eg. `VITE_PUBLIC_TEMPLE_STAKING` and similarly defined variables.

We should have a single, authoritative list, stored in git, of the deployed contract addresses for testnet + mainnet.

This will help ops and post deployment triage.

scooprinder:
I'm working on the helper file in #434 - I can fold this task into that too.

butlerji:
SGTM :)

butlerji:
Closing issue (FYI @scooprinder )

   1. #444 and #433 Cover this off for deprecated/deleted contracts,and actively used contracts
   2. As part of #436, We'll make a single list,
   
If, after that, we want to push the DRYness of the list of contracts, we can revisit.

PR #444, opened 2022-07-12
chore: migrate dapp env vars to constants

nichosystem:
# Description
- fixed protocol local-deploy script
- moved dapp env vars to json constants
- removed some deprecated code from dapp

dependent on #433 

# Checklist


apps/dapp/src/providers/StakingProvider.tsx
```diff
@@ -1,39 +1,14 @@
-import {
-  useState,
-  createContext,
-  useContext,
-  PropsWithChildren,
-} from 'react';
-import { BigNumber, ethers, Signer } from 'ethers';
+import { useState, createContext, useContext, PropsWithChildren } from 'react';
```
nichosystem, line 1: looks like my linter might have a different max length than what this file was previously linted as, sorry for the many changes 🙇

apps/dapp/src/services/MetricsService.ts
```diff
@@ -1,26 +1,23 @@
-//@ts-nocheck
```
nichosystem, line 1: removing this nocheck revealed a ton of errors in this file, so quite a few general fixes besides migrating away from env vars

apps/dapp/src/services/MetricsService.ts
```diff
@@ -83,139 +80,54 @@ export interface SocialMetrics {
...
-      isDevelopmentEnv()
-        ? new ethers.providers.Web3Provider(window.ethereum)
-        : new ethers.providers.AlchemyProvider(
-            ALCHEMY_PROVIDER_NETWORK,
-            ALCHEMY_API_KEY
-          );
-
-    this.signer = new ethers.Wallet(SERVER_PRIVATE_KEY, this.provider);
-
-    this.stableCoinContract = new ERC20__factory(this.signer).attach(
```
nichosystem, line 176: removed `signer` altogether and replaced with the static `Factory.connect(address, provider)` method, since everything should be readonly (we also get to ditch the env var SERVER_PRIVATE_KEY from this 👍 )

protocol/scripts/deploys/helpers.ts
```diff
@@ -63,15 +68,15 @@ export const DEPLOYED_CONTRACTS: {[key: string]: DeployedContracts} = {
...
     // From network/environment
     FRAX: '0x853d955acef822db058eb8505911ed77f175b99e',
-    MULTISIG: "0x4D6175d58C5AceEf30F546C0d5A557efFa53A950",
+    MULTISIG: '0x4D6175d58C5AceEf30F546C0d5A557efFa53A950',
 
     TEMPLE: '0x470ebf5f030ed85fc1ed4c2d36b9dd02e77cf1b7',
     TEMPLE_TEAM_FIXED_PAYMENTS: '0xF7b10A0C780a3906D9A9F3d706EcD2624B6ED84e',
     TEMPLE_TEAM_EPOCH_2: '0xe0Aafcf26576a53Cbec99481607FB53384909C36',
-    TEMPLE_TEAM_EPOCH_3: '', // XXX: FIll in
+    TEMPLE_TEAM_EPOCH_3: '0xf86c2dbd16f05e86bff72ce89b3c2915812e92d0',
```
nichosystem, line 76: only actual change in this file is adding these 2 addresses, everything else is linting

mrfujisawa (approved):

apps/dapp/src/constants/env/local.tsx
```diff
@@ -1,12 +1,36 @@
 import { Environment } from './types';
 
 const env: Environment = {
+  alchemyId: 'VvVv_fBIiRSaTQzL9RQNybD5FSNvtK5c',
+  backendUrl: 'http://localhost:3001',
+  contracts: {
+    exitQueue: '0x0165878A594ca255338adfa4d48449f69242Eb8F',
```
butlerji, line 7: I'd expect this to no longer be used/required by the DApp (IMHO I think this should be tracked as an issue, and hopefully someone in the community can pick it up as a low priority/nice starter task).

apps/dapp/src/constants/env/local.tsx
```diff
@@ -1,12 +1,36 @@
...
+    faith: '0xBEc49fA140aCaA83533fB00A2BB19bDdd0290f25',
```
butlerji, line 8: I'd expect this to no longer be used/required by the DApp (IMHO I think this should be tracked as an issue, and hopefully someone in the community can pick it up as a low priority/nice starter task).

apps/dapp/src/constants/env/local.tsx
```diff
@@ -1,12 +1,36 @@
...
+    farmingWallet: '',
```
butlerji, line 9: What's this used for?

apps/dapp/src/constants/env/local.tsx
```diff
@@ -1,12 +1,36 @@
...
+    fei: '0x8A791620dd6260079BF849Dc5567aDC3F2FdC318',
+    frax: '0x2279B7A0a67DB372996a5FaB50D91eAA73d2eBe6',
+    frax3CrvFarming: '',
+    frax3CrvFarmingRewards: '',
```
butlerji, line 13: What's are these used for?

apps/dapp/src/constants/env/production.tsx
```diff
@@ -1,16 +1,52 @@
 import { Environment } from './types';
 
 const env: Environment = {
```
butlerji, line 3: confirming there are no secrets (and safe to checkin/have in git).

Guessing the dapp doesn't have secrets in general?

apps/dapp/src/constants/env/types.ts
```diff
@@ -3,16 +3,54 @@ interface Subgraphs {
   protocolMetrics: string;
 }
 
-interface Contracts {}
+interface Contracts {
```
butlerji, line 6: Given we have a lovely mono repo @nichosystem - I think this should only live in one place.

I suggest deleting the list of active contracts from the protocol repo, and keeping this as the single source of truth, for active contracts WDYT?

I can leave the markdown of deprecated/deleted contracts in the protocol repo (that feels like the right home for it), but no need to keep 'active' contracts listed in 2 places

butlerji (approved):
LGTM

General comments are more discussion points/things to follow up on. Nothing blocking this merge :)

apps/dapp/src/constants/env/types.ts
butlerji: Update, started a draft PR for this: https://github.com/TempleDAO/temple/pull/446

I'm not going to actually unify the contracts in this PR, Leave it to whoever does issue https://github.com/TempleDAO/temple/issues/435

apps/dapp/src/constants/env/production.tsx
nichosystem: yeah dapp doesnt have secrets

apps/dapp/src/constants/env/local.tsx
nichosystem: currently pulled into the metrics service to display the treasury. I think this work is migrated to the subgraph though, we can return to the metrics service cleanup in a future issue

apps/dapp/src/constants/env/local.tsx
nichosystem: I think we may want to wait a bit until there's some full consensus that faith is for sure dead

PR #446, opened 2022-07-13
Feat/improved deployment

butlerji:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


butlerji:
Worth noting, this is also the first step to allow deployment via github actions.

Which IMHO closes the loop on traceability, and brings us in line with best practice deploy/review for critical code.

protocol/deploys/99-example-deploy-script.ts
```diff
@@ -0,0 +1,42 @@
...
+    CONSTRUCTOR_ARG_1: '';
+    CONSTRUCTOR_ARG_2: '';
+    VESTING_DURATION: '',
+  }
+  ensureExpectedEnvvars(args);
+
+  const [owner] = await ethers.getSigners();
+
+  const someSmartContractFactory = new SomeSmartContract__factory(owner)
+  const someSmartContract: Presale = await deployAndMine("PRESALE", someSmartContractFactory, someSmartContractFactory.deploy,
```
frontier159, line 25: To make it super crisp, I wonder if this can just take one single arg, which is an object of the required `{constructor_arg_name:value}`? Not sure if this is possible with ts magic

```
const someSmartContract: Presale = await deployAndMine(
    "PRESALE",
    someSmartContractFactory,
    someSmartContractFactory.deploy,
    args,
);
```

So the args need to match the constructor kv args (name and value type)

protocol/deploys/README.txt
```diff
@@ -0,0 +1,9 @@
+This directory contains a complete list of all on chain contract deployments
+
+Directory Structure
+  - `mainnet/` all contracts deployed on mainnet
+  - `goreli/` all contracts deployed on goreli testnet
+  - `contract-address.ts` all active deployed contracts (NOTE: to be deleted/unified with dapp)
+  - `helpers.ts` common code to help deployments
+  - `[0-9]+-some-script-name.ts` last migration script run. We try and make the script generic s.t it will run on both
```
frontier159, line 8: nit: Should make this file a markdown

In stax the testnet scripts diverge from mainnet, because there's a bunch of other things we need to do in testnet when the upstream protocol isn't deployed there.

So will probably still need a specific set of `goerli/[0-9]+-some-script-name.ts` scripts. A chunk of them will be the same though.

Similarly with `contract-addresses.ts`, because there are a bunch of addresses specific only to the testnet setup, I'm leaning on creating a testnet specific interface rather than a bunch of empty string/optionals. Can see the gmx branch for a WIP setup of that

protocol/deploys/helpers.ts
```diff
@@ -0,0 +1,94 @@
...
+  console.log('Contract deployed');
+  console.log(`export ${name}=${contract.address}`);
+
+  const deployArgsDir = `${__dirname}/${network.name}`;
+  if (!fs.existsSync(deployArgsDir)) {
+    fs.mkdirSync(deployArgsDir);
+  }
+  fs.writeFileSync(`${deployArgsDir}/${contract.address}.js`, `module.exports = ${JSON.stringify(verifyRenderedArgs, null, 2)}`)
+
+  console.log(`hardhat verify args written to ${deployArgsDir}/${contract.address}.js`);
```
frontier159, line 60: A good step in the right direction for sure! The flow still means that after each contract deploy, they have to go in and update `contract-addresses.ts` manually, one by one.

Think we could enhance by:
1. Pass through the target of this deploy as an arg (rather than just some rando `name` string) eg `TEMPLE_V2_FRAX_PAIR`
2. After it's deployed, update the `contract-addresses.ts` `DEPLOYED_CONTRACTS.${network.name}.${target}`

Less copy/paste messup risk then.

After this is out, I think we can then automate the verify step which would be awesome too (or at least bulk verify?)

protocol/deploys/README.txt
frontier159: Also, if we have changes to a contract over time - eg a v2 which adds extra params to the constructor -- are you thinking that there should be another *new* deploy script so we can keep a fully reproducible history? Or just relying on git history for that?

I guess LEGACY-DEPLOYED.md should capture the v1 contract too

princetonbishop: I agree to the above :)

nichosystem:
stale

PR #448, opened 2022-07-13
Feat/generic distribution contract

butlerji:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist
- [ ] We aren't exposing an attack vector via vesting and multi-claim per account (see @scooprinder comment: https://github.com/TempleDAO/temple/pull/448#discussion_r919686319)


protocol/contracts/admin/MerkleTreeClaim.sol
```diff
@@ -0,0 +1,121 @@
...
+        claimShareMerkleRoot = _claimShareMerkleRoot;
+        totalUnclaimedShares = _totalUnclaimedShares;
+    }
+
+    /**
+     * @notice claim `share` of token `purchaseToken` for this Round. Requires valid `amount` as per `inviteCode`
+     */
+    function claimFor(address account, uint256 share, bytes32[] calldata merkleProof) external {
+        require(account != address(0), "MerkleTreeClaim: Address cannot be 0x0");
+        require(MerkleProof.verify(merkleProof, claimShareMerkleRoot, keccak256(abi.encode(account, share))), "MerkleTreeClaim: Invalid claim");
```
scooprinder, line 62: I know its draft so forgive me if I'm jumping in a little early; just dumping while its on my mind. 

Is this contract intended to be reusable? I.e. to allow multiple claims? If so, there'd need to be some protection here for msg.sender (either as part of the leaf or could use eip712 to verify signature); as after the first time an address has claimed, there is nothing to prevent another address replaying using the proof from the tx data.

butlerji: Posted early to get early feedback - so no need to apologise + please jump in early :)

  1. Each contract is used for one pot of claimable tokens
  2. For any given pot, an account can claim multiple times, mainly to keep claiming during the vesting period
  3. Normally you'd have some kind of address => claim boolean, In order to have vesting, I've got an `address => claimedSoFar`  state to balance the math
  4. because it's claimFor, and the merkle tree is public kowledge anyway, the worst that can happen is I claim and send value to a given address. I can't actually take their tokens.
  
  All untested ofc - but I think your attack vector is impossible (we should think it through again before merging to master! I've added it to the checklist for this PR)

nichosystem:
stale & replaced by #628

PR #482, opened 2022-08-02
Ascend follow ups

mrfujisawa:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist

### DIFF
--- a/apps/dapp/src/components/Pages/Ascend/components/AuctionContext/index.tsx
+++ b/apps/dapp/src/components/Pages/Ascend/components/AuctionContext/index.tsx
@@ -72,7 +72,7 @@ interface Props {
 
 export const AuctionContextProvider: FC<Props> = ({ pool, children }) => {
   const { wallet } = useWallet();
-  const [buy, sell] = pool.tokens;
+  const [sell, buy] = pool.tokens;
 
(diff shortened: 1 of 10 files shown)


apps/dapp/src/components/Pages/Ascend/components/AuctionContext/index.tsx
```diff
@@ -72,22 +72,22 @@ interface Props {
 
 export const AuctionContextProvider: FC<Props> = ({ pool, children }) => {
   const { wallet } = useWallet();
-  const [buy, sell] = pool.tokens;
+  const [sell, buy] = pool.tokens;
```
scooprinder, line 75: is this due to the ordering of tokens in goerli? Are the mainnet tokens going to be in the same alphanumeric order?

mrfujisawa: good point, ill fix this.

mrfujisawa: i've been assuming that the tokens would be sorted from the graph but probably not reasonable to assume. updating that list as well

PR #485, opened 2022-08-04
fix: deprecate ogt staking UI and routes

zelkibuilds:
# Description
What does this PR solve?
There have been calls to deprecate OGT staking from the dapp since users are still staking their temple when we offer 0% APY and it's leading to some confusion. This PR removes both the stake page/modal and its associated routes.
There's also been added a fallback route that maps anything that's not `/dapp/trade/unstake` to `/dapp/trade` as `/unstake` and `/` are the only valid routes under `/dapp/trade` if this PR gets merged. This is currently an issue in prod as well. You can see the unintended results of accessing an unmapped url like [https://templedao.link/dapp/trade/unmapped](https://templedao.link/dapp/trade/unmapped)

# Checklist


mrfujisawa (approved):

scooprinder:
any reason why we can't merge this?

PR #486, opened 2022-08-04
Posthog integration

marshall2112:
# Description

Initial integration with posthog, https://posthog.com/docs/integrate/client/js

# Checklist

### DIFF
--- a/apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
+++ b/apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
@@ -29,10 +29,10 @@ export const useDepositToVault = (vaultContractAddress: string, onSuccess?: Call
   const [getZappedAssetValue] = useGetZappedAssetValue();
 
   const { openNotification } = useNotification();
-  const { captureEvent } = AnalyticsService.getInstance();
+  const { captureEvent } = AnalyticsService;
 
   const handler = async (ticker: TICKER_SYMBOL, amount: string, useFaith = false) => {
-    captureEvent('[vault-deposit]', { amount });
+    captureEvent('vault-deposit', { amount, vaultContractAddress });
     if (!signer || !wallet) {
       console.error(`
         Attempted to deposit to vault: ${vaultContractAddress} without a valid signer or wallet address.
--- a/apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
+++ b/apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
@@ -13,6 +13,7 @@ import { getBigNumberFromString } from 'components/Vault/utils';
 import { ZERO } from 'utils/bigNumber';
 import { createTokenFactoryInstance } from './use-token-vault-proxy-allowance';
 import { formatJoiningFee } from 'components/Vault/utils';
+import { AnalyticsService } from 'services/AnalyticsService';
 
 import env from 'constants/env';
 
@@ -28,8 +29,10 @@ export const useDepositToVault = (vaultContractAddress: string, onSuccess?: Call
   const [getZappedAssetValue] = useGetZappedAssetValue();
 
   const { openNotification } = useNotification();
+  const { captureEvent } = AnalyticsService.getInstance();
 
   const handler = async (ticker: TICKER_SYMBOL, amount: string, useFaith = false) => {
+    captureEvent('[vault-deposit]', { amount });
     if (!signer || !wallet) {
       console.error(`
         Attempted to deposit to vault: ${vaultContractAddress} without a valid signer or wallet address.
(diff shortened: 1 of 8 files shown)


0xalecks:
Deploy preview for _dapp_ ready!

✅ Preview
https://dapp-6cmpijmvk-templedao.vercel.app
https://pr-486-preview.templedao.link

Built with commit 751a5aeed410133d37be240cfc2333ada5337e7a.

mrfujisawa:
Instead of including this script and initializing globally in the document I think we should try to use the npm package and initialize it inside our javascript code

mrfujisawa:
Screen Shot 2022-08-04 at 3 46 30 PM

The reason being is that if we ever want to call methods on the Posthog instance it'll just be cleaner to directly access the posthog instance instead of reaching into the global scope

marshall2112:
> Instead of including this script and initializing globally in the document I think we should try to use the npm package and initialize it inside our javascript code

@mrfujisawa https://github.com/TempleDAO/temple/pull/486/commits/2ebb5cd015d6ee6341507986f42f1101b1c229d1

apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
```diff
@@ -28,8 +29,10 @@ export const useDepositToVault = (vaultContractAddress: string, onSuccess?: Call
   const [getZappedAssetValue] = useGetZappedAssetValue();
 
   const { openNotification } = useNotification();
+  const { captureEvent } = AnalyticsService.getInstance();
 
   const handler = async (ticker: TICKER_SYMBOL, amount: string, useFaith = false) => {
+    captureEvent('[vault-deposit]', { amount });
```
0xalecks, line 35: can we add vault name too, just to future proof for later when we have more vaults

apps/dapp/src/services/AnalyticsService.ts
```diff
@@ -0,0 +1,35 @@
+import env from 'constants/env';
+import posthog from 'posthog-js';
+import { isDevelopmentEnv } from 'utils/helpers';
+
+export class AnalyticsService {
```
0xalecks, line 5: Why don't we export the instance directly? Currently, every time you want to use it, you have to import the service, then get the instance. It would be simpler to do something like :

```
 import { analytics } from 'services/AnalyticsService';
 
 analytics.init()
 analytics.captureEvent()

```

apps/dapp/src/services/AnalyticsService.ts
```diff
@@ -0,0 +1,35 @@
...
+  private static instance: AnalyticsService;
+  private static shouldSendAnalytics: boolean;
+
+  private constructor() {
+    AnalyticsService.shouldSendAnalytics = !isDevelopmentEnv();
+  }
+
+  public static getInstance(): AnalyticsService {
+    if (!AnalyticsService.instance) {
+        AnalyticsService.instance = new AnalyticsService();
```
0xalecks, line 15: Is there a need to create an instance vs having everything be static?

apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
0xalecks: also, why the brackets? is that some convention in posthog?

0xalecks (changes requested):

apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
marshall2112: Yes will add the vault name.

Bracket? I _thought_ it was convention, but think I was mistaken. I'll remove them.

apps/dapp/src/services/AnalyticsService.ts
marshall2112: Yes I'll simplify it.

apps/dapp/src/services/AnalyticsService.ts
marshall2112: I'll ditch the singleton.

apps/dapp/src/hooks/core/use-deposit-to-vault.tsx
marshall2112: We can add vault address @0xalecks will that work?

marshall2112: https://github.com/TempleDAO/temple/pull/486/commits/59ec0703e6bc95876708be100bdcdef4e74ceb5d

apps/dapp/src/services/AnalyticsService.ts
marshall2112: https://github.com/TempleDAO/temple/pull/486/commits/59ec0703e6bc95876708be100bdcdef4e74ceb5d

apps/dapp/src/services/AnalyticsService.ts
marshall2112: https://github.com/TempleDAO/temple/pull/486/commits/59ec0703e6bc95876708be100bdcdef4e74ceb5d

apps/dapp/src/main.tsx
```diff
@@ -68,6 +69,9 @@ if (env.sentry) {
   });
 }
 
+const { init } = AnalyticsService;
+init();
```
0xalecks, line 73: Small nit, it doesn't matter here cause it's a static object but doing it this way, instead of just `AnalyticsService.init()` can be problematic. If you "pull off" the method like you did, it loses its scope and has no access to `this`

0xalecks (approved):

PR #491, opened 2022-08-05
Improve captureEvent

marshall2112:
# Description

Capture event moved to the success callback. Also edited the success callback to pass ticker and amount, and now pass name (vault.id) to posthog.

# Checklist

### DIFF
--- a/apps/dapp/src/components/Pages/Core/VaultPages/Stake.tsx
+++ b/apps/dapp/src/components/Pages/Core/VaultPages/Stake.tsx
@@ -22,6 +22,7 @@ import { useGetZappedAssetValue } from 'hooks/core/use-get-zapped-asset-value';
 import EllipsisLoader from 'components/EllipsisLoader';
 import { ZERO } from 'utils/bigNumber';
 import { getBigNumberFromString, formatBigNumber, formatJoiningFee } from 'components/Vault/utils';
+import { AnalyticsService } from 'services/AnalyticsService';
 
 export const Stake = () => {
   const { activeVault: vault } = useVaultContext();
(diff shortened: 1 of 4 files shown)


apps/dapp/src/components/Pages/Core/VaultPages/Stake.tsx
```diff
@@ -48,10 +46,14 @@ export const Stake = () => {
 
   const [_, refreshBalance] = useVaultBalance(vault.id);
   const [{ isLoading: refreshIsLoading }, refreshWalletState] = useRefreshWalletState();
-  const [deposit, { isLoading: depositLoading, error: depositError }] = useDepositToVault(vault.id, async () => {
-    refreshBalance();
-    refreshWalletState();
-  });
+  const [deposit, { isLoading: depositLoading, error: depositError }] = useDepositToVault(
+    vault.id,
+    async (ticker, amount) => {
```
0xalecks, line 51: Aren't `ticker` and `amount` available on this page? Not sure if you really needed to pass them from the hook here, but it's fine I guess

0xalecks (approved):

apps/dapp/src/components/Pages/Core/VaultPages/Stake.tsx
marshall2112: Technically yes... but my thinking is that they could change between the button click, and the transaction finishing. E.g. user changes the amount after submitting the txn, and we end up with the wrong amount in this success callback. So IMO it's easy enough to pass it in this way so we know exactly the amount/ticker and not rely on the existing data that's subject to change based on user input.

PR #497, opened 2022-08-09
Change INITIAL_STATE of isFraxSellDisabled from true to false

### LINKED ISSUE #475: Selling $temple doesn't always show $frax as an option, even when above IV
Steps to reproduce:
1. Navigate to https://templedao.link/dapp/trade directly 
2. Click arrow so we're selling Temple
3. Only FEI is available to sell into, even when above IV

Workaround:
1. When you type any value into the Temple input, this then allows user to select between $frax and $fei

Expected behaviour:
When above IV, the option to sell into Frax or Fei should appear without user interaction.

BoxedFruits:
# Description
What does this PR solve?

When first switching to selling $TEMPLE, $FRAX isn't always show as an option in the input select. This was being caused by _INITIAL_STATE_ defaulting _isFraxSellDisabled_ to false, causing the [_buildSelectConfig_ ](https://github.com/TempleDAO/temple/blob/stage/apps/dapp/src/components/Pages/Core/Trade/utils.ts#L32-L37)to render only the $FEI ticker rather than a select. 

Fixes #475 

# Checklist

### DIFF
--- a/apps/dapp/src/components/Pages/Core/Trade/constants.ts
+++ b/apps/dapp/src/components/Pages/Core/Trade/constants.ts
@@ -29,7 +29,7 @@ export const INITIAL_STATE: SwapReducerState = {
   buttonLabel: createButtonLabel(TICKER_SYMBOL.FRAX, TICKER_SYMBOL.TEMPLE_TOKEN, SwapMode.Buy),
   isTransactionPending: false,
   isSlippageTooHigh: false,
-  isFraxSellDisabled: true,
+  isFraxSellDisabled: false,
   error: null,
 };
 


scooprinder:
Suspect because it maybe because secrets are not bought across when creating a PR from an external fork - will look into this. The important thing is the dapp still builds.

scooprinder:
Appreciate its a small change but will use this PR as a guinea pig to iron out any kinks in the build pipeline

scooprinder:
So it looks like we're unable to preview-deploy this due to https://github.com/orgs/community/discussions/26242 

Given the small change in scope, are you ok for this to be merged @0xalecks and we can test on stage?

0xalecks (approved):

PR #501, opened 2022-08-10
Add LBP swap event

marshall2112:
# Description

https://discord.com/channels/847178511604252741/989472996886474773/1006652918461640736

Add event for LBP swap

# Checklist

### DIFF
--- a/apps/dapp/src/components/Pages/Ascend/components/Trade/hooks/use-vault-trade-state.ts
+++ b/apps/dapp/src/components/Pages/Ascend/components/Trade/hooks/use-vault-trade-state.ts
@@ -8,7 +8,8 @@ import { useNotification } from 'providers/NotificationProvider';
 import { useVaultContract } from './use-vault-contract';
 import { getSwapLimit, getSwapDeadline } from '../utils';
 import { useAuctionContext } from '../../AuctionContext';
-import { getBalancerErrorMessage } from 'utils/balancer'
+import { getBalancerErrorMessage } from 'utils/balancer';
+import { AnalyticsService } from 'services/AnalyticsService';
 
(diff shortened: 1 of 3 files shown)


apps/dapp/src/components/Pages/Ascend/components/Trade/hooks/use-vault-trade-state.ts
```diff
@@ -249,9 +250,16 @@ export const useVaultTradeState = (pool: Pool) => {
         title: 'Swap success',
         hash: transaction.hash,
       });
+
+      AnalyticsService.captureEvent('lbp-swap', {
```
mrfujisawa, line 254: Does it make sense to create an events file somewhere where we can define all these events?  It might also be useful to name space the events...
could do something like
```
export enum LBPEvent {
   Swap = 'Swap',
  ... other events specific to LBP stuff
}
```

it might be overkill but could define lbp event functions so the expected event attributes are more clearly defined as well----
for instance
````
const logLBPSwapEvent = (tokenSold: string, tokenBought: string, amount: string, poolId: string) => {
   AnalyticsService.captureEvent(LBPEvent.Swap, { tokenSold, tokenBought, amount, poolId);
}
```
might be thinking ahead a bit, but if we are logging similar events in different places maybe could be useful

marshall2112: OK as a start, I moved the event names to a constants file https://github.com/TempleDAO/temple/pull/501/commits/947290234b720d1f8448e595c22695e2f6504881.

Can discuss more in Discord, my feeling is that we should keep the engineering to a minimum unless/until we find out how many custom events we might want. Asked Doc in chat.

scooprinder (approved):
LGTM

PR #545, opened 2022-09-06
update intervals

0xalecks:


apps/dapp/src/components/Pages/Ascend/components/Trade/hooks/use-vault-trade-state.ts
```diff
@@ -183,7 +183,7 @@ const INITIAL_QUOTE_STATE = {
   error: null,
 };
 
-const QUOTE_INTERVAL = 7500; // 7.5  seconds
+const QUOTE_INTERVAL = 1500; // 7.5  seconds
```
scooprinder, line 186: Should this be `15000`?

scooprinder (approved):
Lets update the comment in `use-vault-trade-state.ts` when we refactor this later.

PR #581, opened 2022-10-14
github workflow to use slither

mountainpath9:


frontier159 (approved):
LGTM

.github/workflows/slither.yaml
```diff
@@ -0,0 +1,32 @@
+# We manually setup and run slither here, as the prebaked github action
+# (at https://github.com/crytic/slither-action)
+# fails with permission errors inside the docker container for this project
+
+name: Slither Analysis
+on: [push]
```
nichosystem, line 6: add a paths to this trigger so that it only runs on "protocol/**" (see https://github.com/TempleDAO/temple/blob/stage/.github/workflows/dapp-prod.yaml#L2)

frontier159: done

nichosystem (approved):

PR #593, opened 2022-10-21
Governance contracts for templar NFTs and elder elections

mountainpath9:


protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        if (candidates[discordId]) {
+          candidates[discordId] = false;
+          emit UpdateNomination(discordId, candidates[discordId]);
+        }
+    }
+
+    function setEndorsements(uint256[] memory discordIds) external {
+        _setEndorsements(msg.sender, discordIds);
```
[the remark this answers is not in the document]
frontier159: Where did we land on the mechanics of vote weight? Does an initiate's endorsements have more weight than an acolyte's endorsements?

If so, how do we link the voter's account (might be team, might be anon/community)? Does this merke tree also need to have the account?
```
    /// @notice root hash of off-chain merkle tree recording
    /// temple roles ie :(discordId, maxTempleRole)
    bytes32 public templarRolesMerkleRoot;
 ```
 
So then we can verify that if the voting templar is team (and also verify their role at the time)

File: protocol/contracts/governance/ElderElection.sol [hunk 2]
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    function _setEndorsements(address account, uint256[] memory discordIds) internal {
+        endorsementsBy[account] = discordIds;
```
frontier159, line 123: docs recommend to use `uint256[] calldata discordIds`
https://docs.soliditylang.org/en/v0.8.4/types.html?highlight=calldata#data-location

I wonder if we should put a bound on the number of allowed endorsements (eg 10?), just so this storage isn't ridiculous?

Probably need a custom getter to get all discordIds back in one hit. With the default getter, caller would need to provide an array index to get 1-by-1 (and so would need to know then length of the array)
```
function getEndorsementsBy(address account) external returns (uint256[] calldata) {
    return endorsementsBy[account];
}
```

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @notice Toggle the 
+     * (assuming the owner has given authority for the caller to act on their behalf)
+     *
+     * @dev amount is explicit, to allow use case of partial vault withdrawals
+     */
+    // TODO(butlerji): Need to test this works. The idea is I'd sign a message on mainnet, which gets sent to our backend, then sent to 
+    //                 the mempool on whatever L2 we are running this on.
+    function relayedSetEndorsementsFor(address account, uint256[] memory discordIds, uint256 deadline, bytes calldata signature) public {
```
frontier159, line 135: Can be external, no need for private here?

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * @notice Change the nomination fee
+     * 
+     * Designed to be low, but a defense in depth to stop peeps putting up a nomination
+     * for the lolz.
+     */
+    function setNominationFee(uint256 _nominationFee) external onlyOwner {
+        nominationFee = _nominationFee;
```
frontier159, line 68: Missing events for setters?

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,78 @@
...
+        if (_exists(dclaim.discordId)) {
+            _transfer(ownerOf(dclaim.discordId), account, dclaim.discordId);
+
+        } else {
+            _safeMint(account, dclaim.discordId);
+        }
+    }
+
+    struct DiscordClaim {
+        uint256 discordId;
```
frontier159, line 59: Out of interest, why is this a struct? Could just unwrap it since it's only a uint256?

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    function hashDiscordClaim(DiscordClaim memory dclaim) private pure returns (bytes32) {
+        return keccak256(abi.encode(
+            DISCORD_CLAIM_TYPEHASH,
+            dclaim.discordId
+        ));
+    }
+
+    function tokenURI(uint256 discordId) public pure override returns (string memory) {
+        return string.concat("https://discordapp.com/users/", Strings.toString(discordId));
```
frontier159, line 73: This doesn't compile atm...I got it compiling with:
```

    function tokenURI(uint256 discordId) public pure override returns (string memory) {
        return string(
            bytes.concat("https://discordapp.com/users/",
            bytes(Strings.toString(discordId)))
        );
    }
```
But a bit gross, may be another way

nichosystem, line 73: pretty sure string concat is not available in solidity 0.8.4

nichosystem: also will need to import '@openzeppelin/contracts/utils/Strings.sol';

clusby: Yeh only from 0.8.12+

Pretty sure strings.sol is pulled in as a transient dependency, so not absolutely necessary but probably good to do regardless

protocol/contracts/governance/ElderElection.sol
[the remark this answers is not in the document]
mountainpath9: This PR reflects my understanding of the how vote weights are used:

https://github.com/TempleDAO/shrine/pull/134

An address endorses candidates. The candidates will get vote points from the endorsement:

- if the address holds staked temple, then it contributes token vote points scaled according to the token balance.
- if the address holds the NFT for a temple role, then it contributes team vote points, scaled according to the discord role.

both of the above can be true for a given address.

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @notice Nominate the given discord ID a candidate.
+     * @dev The given address must hold the NFT for the specified discord id
+     */
+    function nominate(
+        address account,
+        uint256 discordId
+    ) external {
+        if (templars.ownerOf(discordId) != account) {
```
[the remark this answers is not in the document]
mountainpath9: Yep - will fix.

protocol/contracts/governance/ElderElection.sol
mountainpath9: Yes - will limit endorsements to active candidates.

protocol/contracts/governance/ElderElection.sol
mountainpath9: which "private" ?

frontier159: sorry - i meant 'public', updated

protocol/contracts/governance/Templar.sol
mountainpath9: Probably doesn't need to be a struct but:
  - I thought it might need more fields.
  - I was trying to follow EIP-712 conventions

protocol/contracts/governance/ElderElection.sol
```diff
@@ -66,23 +69,29 @@ contract ElderElection is Ownable, EIP712 {
...
+        emit UpdateNominationFee(_nominationFee);
     }
 
     /**
-     * @notice Update the merkle tree root for temple roles.
+     * @notice Update the hash root for temple roles.
      */
-    function setTemplarRolesMerklRoot(bytes32 _templarRolesMerkleRoot) external onlyOwner {
-        templarRolesMerkleRoot = _templarRolesMerkleRoot;
+    function setTemplarRolesMerklRoot(bytes32 _templarRolesHash) external onlyOwner {
```
frontier159, line 78: can rename `setTemplarRolesMerklRoot()` too

protocol/contracts/governance/ElderElection.sol
```diff
@@ -93,24 +102,32 @@ contract ElderElection is Ownable, EIP712 {
               nominationFeeToken.safeTransferFrom(account, address(this), nominationFee);
           }
           candidates[discordId] = true;
+          numCandidates += 1;
           emit UpdateNomination(discordId, candidates[discordId]);
```
frontier159, line 106: I would personally just hardcode to `true` rather than having another storage lookup
`emit UpdateNomination(discordId, true);`

protocol/contracts/governance/ElderElection.sol
```diff
@@ -66,23 +69,29 @@ contract ElderElection is Ownable, EIP712 {
...
+        emit UpdateTemplarRolesHash(_templarRolesHash);
     }
 
     /**
      * @notice Nominate the given discord ID a candidate.
-     * @dev The given address must hold the NFT for the specified discord id
+     * @dev The sender must hold the NFT for the specified discord id
      */
-    function nominate(
+    function nominate(uint256 discordId) external {
```
frontier159, line 87: Just confirming:
1. Only a self-nomination is permitted? (Which is fine, just for my understanding)
2. Anyone that wants to be nominated needs to transact directly on the L2 - no relayer for this.

Also imo we don't need the separate internal function as it just means more gas for no benefit.

protocol/contracts/governance/ElderElection.sol
```diff
@@ -14,20 +14,23 @@ import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
...
  */
 contract ElderElection is Ownable, EIP712 {
     using Counters for Counters.Counter;
     using SafeERC20 for IERC20;
 
-    /// @notice The NFT ids of the Templars thar are candidates for the ElderElection
+    /// @notice The NFT ids of the Templars that are candidates for the ElderElection
     mapping(uint256 => bool) public candidates;
 
+    /// @notice The number of actives candidates
```
frontier159, line 31: nit: minor typo `active`

protocol/contracts/governance/ElderElection.sol
```diff
@@ -14,20 +14,23 @@ import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
...
+ * who hold staked temple or who are temple team members have vote value.
  *
  * EIP712 structured data signing supports 'gasless voting' via a relay
  */
 contract ElderElection is Ownable, EIP712 {
     using Counters for Counters.Counter;
     using SafeERC20 for IERC20;
 
-    /// @notice The NFT ids of the Templars thar are candidates for the ElderElection
+    /// @notice The NFT ids of the Templars that are candidates for the ElderElection
```
frontier159, line 28: may be worth mentioning in the docstring that the key is the discordId

protocol/contracts/governance/ElderElection.sol
```diff
@@ -120,6 +137,10 @@ contract ElderElection is Ownable, EIP712 {
     }
 
     function _setEndorsements(address account, uint256[] memory discordIds) internal {
+        if (discordIds.length > numCandidates) {
+          revert TooManyEndorsements();
+        }
+
         endorsementsBy[account] = discordIds;
```
frontier159, line 144: Is it fair to say it's the responsibility of the caller to ensure `discordIds` are all valid - the contract won't check that `candidates[discordId] == true` for each of these ids (this could be a large list - so don't want to check that on chain)

protocol/contracts/governance/ElderElection.sol
mountainpath9: It's a sad programming language that do that optimisation for itself....

protocol/contracts/governance/ElderElection.sol
mountainpath9: Yep - there's no harm in the discordIds being invalid. They just won't count for anything.

protocol/contracts/governance/Templar.sol
frontier159: Ah interesting on the EIP-712. I haven't used the signing in anger before, good to know thanks

protocol/contracts/governance/ElderElection.sol
mountainpath9: Tidied up. The bot account will have the `CAN_NOMINATE` role on the EldersElection.sol contract, and the `CAN_ASSIGN` role on the Templars.sol NFT.

protocol/contracts/governance/ElderElection.sol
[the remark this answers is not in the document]
mountainpath9: The current best view of the voting process is in the MVP doc. Let's have discussions about the details there.

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,149 @@
...
+        templars.checkExists(discordId);
+
+        if (candidates[discordId]) {
+          candidates[discordId] = false;
+          numCandidates -= 1;
+          emit UpdateNomination(discordId, false);
+        }
+    }
+
+    function setEndorsements(uint256[] memory discordIds) external {
```
frontier159, line 84: Best practice to use `calldata` instead of `memory` for param args if you can

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,149 @@
...
+        if (discordIds.length > numCandidates) {
+          revert TooManyEndorsements();
+        }
+
+        endorsementsBy[account] = discordIds;
+        emit UpdateEndorsements(account, discordIds);
+    }
+
+    /**
+     * @notice Toggle the
```
frontier159, line 98: Need to update this comment and remove the butlerji comment

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,149 @@
...
+     * "Consume a nonce": return the current value and increment.
+     */
+    function _useNonce(address _owner) internal returns (uint256 current) {
+        Counters.Counter storage nonce = _nonces[_owner];
+        current = nonce.current();
+        nonce.increment();
+    }
+
+    event UpdateNomination(uint256 discordId, bool isNominated);
+    event UpdateEndorsements(address account, uint256[] discordId);
```
frontier159, line 137: Should add `indexed` keyword to the params that are likely to be searched on. Basically allows those params to be used as filters.

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,149 @@
...
+ * The election runs "continuously" - timing for when the results are actually
+ * actioned is handled at the social/policy layer.
+ * 
+ * Any address can vote by endorsing candidates. However only endorsements from voters
+ * who hold temple (aka "token stakeholders") or who are temple team members (aka "impact
+ * stakeholders") have vote value.
+ *
+ * EIP712 structured data signing supports 'gasless voting' via a relay
+ */
+contract ElderElection is EIP712, Ownable, AccessControl {
```
frontier159, line 27: I wonder if we should add Pausable contract to this, such that (if need be) we can pause the election voting when we want to snapshot at the policy layer?

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+            ERC721.supportsInterface(interfaceId) || 
+            AccessControl.supportsInterface(interfaceId)
+        );
+    }
+
+    /**
+     *  Update the base URI for tokens.
+     */
+    function setBaseUri(string calldata _baseUri) external onlyOwner {
+        baseUri = _baseUri;
```
frontier159, line 37: add event

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            _safeMint(account, discordId);
+        }
+
+        if (keccak256(abi.encodePacked(templeRole[discordId])) != keccak256(abi.encodePacked(_templeRole))) {
+            templeRole[discordId] = _templeRole;
+            emit UpdateTempleRole(discordId, _templeRole);
+        }
+    }
+
```
frontier159, line 70: Do you think we'll ever need to burn? Or is transferring to another account (eg a dao owned 'expired ids' multisig enough?)

Or do we assume that users will always have control, and the nfts may always exist over time -- just that it has no impact because there'll be no votes / they're not nominated...?

(Sorry I may have answered this before)

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+    function checkExists(uint256 discordId) public view {
+        if (!_exists(discordId)) {
+            revert InvalidTemplar(discordId);
+        }
+    }
+
+    function getTempleRole(uint256 discordId) public view returns (string memory) {
```
frontier159, line 77: I think this getter exists by default, no?

`templeRole(discordId) => string`

protocol/contracts/governance/Templar.sol
mountainpath9: It's not clear to me what value an event would add here. How do you see it being used?

protocol/contracts/governance/Templar.sol
mountainpath9: > Or do we assume that users will always have control, and the nfts may always exist over time -- just that it has no impact because there'll be no votes / they're not nominated...?

That's what I assumed.

protocol/contracts/governance/Templar.sol
frontier159: Best practice for all protocol set style events to emit events such they can be responded too off-chain. slither type tools will suggest so too

In this case, hypothetically, it would allow something like a metadata sync to be triggered for all nfts in the collection.

mountainpath9: OK - I'll add it. I though this was only so that tools like metamask coujld show a link.

protocol/contracts/governance/ElderElection.sol
mountainpath9: Given that reading votes etc requires processing the emitted events, I thought we would just read the events up to a given time to get the result at that time. 

Happy to add pausing if needed.

protocol/contracts/governance/Templar.sol
mountainpath9: Will remove.

protocol/test/templar-nft-tests.ts
```diff
@@ -0,0 +1,87 @@
...
+      const templar = TEMPLAR.connect(ben);
+      await expect(templar.assign(amandaAddress, 1000, "acolyte"))
+        .to.be.revertedWith("AccessControl:");
+    }
+  });
+
+  it("Assignment mints an NFT", async () => {
+
+    const amandaAddress: string = await amanda.getAddress();
+
```
frontier159, line 58: Check for `InvalidAddress`?

protocol/test/templar-nft-tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    const DISCORD_ID = 1000;
+    const TEMPLE_ROLE = "acolyte";
+    const templar = TEMPLAR.connect(assigner);
+    await (await templar.assign(amandaAddress, DISCORD_ID, TEMPLE_ROLE)).wait();
```
frontier159, line 62: Don't need the extra `wait()`

should check the event gets emitted

protocol/test/templar-nft-tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    const templar = TEMPLAR.connect(assigner);
+
+    await (await templar.assign(amandaAddress, DISCORD_ID, TEMPLE_ROLE)).wait();
+
+    await (await templar.assign(benAddress, DISCORD_ID, "initiate")).wait();
+
+    expect(await TEMPLAR.ownerOf(DISCORD_ID)).to.eq(benAddress);
+    expect(await TEMPLAR.templeRole(DISCORD_ID)).to.eq("initiate");
+    expect(await TEMPLAR.tokenURI(DISCORD_ID)).to.eq("https://discordapp.com/users/1000");
+  });
```
frontier159, line 86: Should probably add a test for `checkExists()`, and also that the constructor sets roles correctly?

protocol/test/templar-nft-tests.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it("Only configured role can assign", async () => {
+
+    const amandaAddress: string = await amanda.getAddress();
+
+    // The assigner can assign NFTs
+    {
+        const templar = TEMPLAR.connect(assigner);
+        await templar.assign(amandaAddress, 1000, "acolyte");
```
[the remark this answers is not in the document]
mountainpath9: > a later version of hardhat takes either the Promise<T> or T for contract functions

Uggh. I don't like this all - yet more js typeless magic.

`Promise<T>` and `T` are different types, and it's `awaits` job to convert from `Promise<T>` to T.

protocol/test/templar-nft-tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    const assignerAddress: string = await assigner.getAddress();
+    const canAssignRole: string  = await TEMPLAR.CAN_ASSIGN();
+    await (await TEMPLAR.grantRole(canAssignRole, assignerAddress)).wait();
+
+  })
+
+  it("Only owner can setBaseUri", async () => {
+
+    // Owner can
+    await TEMPLAR.setBaseUri("https://new-base-uri/");
```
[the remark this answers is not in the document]
mountainpath9: 2/ check the getter() after you set the uri to make sure it matches

There's a separate test for the effect of setting the URI later.

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+        if (account == address(0)) revert InvalidAddress(account);
+
+        if (_exists(discordId)) {
+            if (ownerOf(discordId) != account) {
+                _transfer(ownerOf(discordId), account, discordId);
+            }
+        } else {
+            _safeMint(account, discordId);
+        }
+
```
medariox, line 67: Would add an event here e.g.:
```
Assigned(ownerOf(discordId), account, discordId, _templeRole)
```

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+        if (keccak256(abi.encodePacked(templeRole[discordId])) != keccak256(abi.encodePacked(_templeRole))) {
```
medariox, line 68: Consider adding a new function e.g. `updateTempleRole(discordId, _templeRole)`, because we might want to set a role without having to mint or transfer.

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
+//SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
```
nichosystem, line 6: don't think we're using Strings now

protocol/scripts/deploys/governance/contract-addresses.ts
```diff
@@ -0,0 +1,38 @@
+import { network } from "hardhat";
+
+export interface DeployedContracts {
+    MULTISIG: string;
+    TEMPLAR_NFT: string;
+    ELDER_ELECTION: string;
+}
+
+export const DEPLOYED_CONTRACTS: { [key: string]: DeployedContracts } = {
```
nichosystem, line 9: It might be nice to have these in the same file as the other addresses (deploys/helpers.ts), everything in one place

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,143 @@
+//SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import "@openzeppelin/contracts/utils/Counters.sol";
+import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
```
nichosystem, line 8: dont need merkle

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,143 @@
...
+    uint256 public numCandidates;
+
+    /// @notice endorsements by a given voter
+    mapping(address => uint256[]) public endorsementsBy;
+
+    /// @notice the NFT contract for templars
+    Templar public templars;
+
+    /// @notice Nonces used in relayed voting requests
+    mapping(address => Counters.Counter) public _nonces;
```
nichosystem, line 47: why name with underscore (_nonces) if it's public? thought that was primarily for private vars

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,143 @@
...
+        emit UpdateEndorsements(account, discordIds);
+    }
+
+    /**
+     * @notice Set the endorsements for an account via relayed, signed request. 
+     */
+    function relayedSetEndorsementsFor(address account, uint256[] calldata discordIds, uint256 deadline, bytes calldata signature) external {
+        if (block.timestamp > deadline) revert DeadlineExpired(deadline);
+
+        bytes32 structHash = keccak256(abi.encode(SET_ENDORSEMENTS_FOR_TYPEHASH, account, discordIds, deadline, _useNonce(account)));
```
nichosystem, line 103: If the signature has a deadline, is there any value in including a nonce? (I would like to not have to fetch the current nonce before signing on the FE)

protocol/contracts/governance/Templar.sol
nichosystem: cleaner to leave mint/transfer since they emit Transfer already, and UpdateTempleRole is emitted separately?

protocol/contracts/governance/Templar.sol
nichosystem: @pocin wdyt?

protocol/contracts/governance/Templar.sol
medariox: Makes it much harder to get all the needed info - we have to listen to 3 separate events instead of just 1.

mountainpath9: I didn't realise that relying on multiple events would make the subgraph implementation harder.

Is it normal to add new events to contracts in order to met subgraph needs?

protocol/contracts/governance/Templar.sol
mountainpath9: It was a conscious decision to have a single function here, used for minting/transfering nfts and updating the metadata here. No mint or transfer will happen unless it is necessary.

Does this make the client side harder?

protocol/contracts/governance/ElderElection.sol
mountainpath9: @butlerji this was from your original implementation. Is the nonce necessary?

protocol/contracts/governance/Templar.sol
```diff
@@ -0,0 +1,89 @@
...
+ */
+contract Templar is ERC721, AccessControl {
+
+    bytes32 public constant OWNER = keccak256("OWNER");
+    bytes32 public constant CAN_ASSIGN = keccak256("CAN_ASSIGN");
+
+    string public baseUri = "https://discordapp.com/users/";
+
+     // Mapping from discordId ID to max temple role
+    mapping(uint256 => string) public templeRole;
```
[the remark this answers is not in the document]
mountainpath9: I considered that, however I decide to go with strings so that the contract will continue to work when roles are changed 

This is will happen as "master" is removed and "higher initiate" is added. I feel like roles are likely to change further as gov is refined.

protocol/contracts/governance/ElderElection.sol
mountainpath9: @butlerji Does _nonces need to be public?

protocol/scripts/deploys/governance/contract-addresses.ts
mountainpath9: @frontier159 thoughts? I'm new to the structure of the deploy scripts etc.

protocol/contracts/governance/Templar.sol
medariox: Why not? After all, `assign()` is not the same as `transfer()/mint()`. If the event provides value, it should be added. I would even argue if the correct way is to have events for transfer and mint. Do they provide useful information for our use case? We don't actually care if the token was transferred or minted. Always add events for the sepcific use case.

protocol/contracts/governance/ElderElection.sol
butlerji: Happy for nonce to be public.

My 'default' is to make things public as it makes debugging easier, but use _ as a convention so other contributors know the _intent_ is to be private.

However, given the caller needs the next nonce, happy to drop the underscore and make this public :)

protocol/contracts/governance/ElderElection.sol
butlerji: Defence in depth on my part. 

Timeout stops replay attacks if the tx fails (not perfect, but it's the best we can do)

Nonce stops replay attacks if the Tx is successful.

Mild preference to leave both for the meta reason others will copy/paste this code and use it in other contexts (and also, because it's all public, to stop any mis-guided FUD).

It is mild however, so happy to have just timeout if it makes it materially less effort for ya'll

protocol/contracts/governance/Templar.sol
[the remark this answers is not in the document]
butlerji: Strings for now, 100% roles will change like crazy :p

butlerji (approved):
LGTM :)

protocol/scripts/deploys/governance/contract-addresses.ts
frontier159: The reason I split it out is because there would be a tonne of empty strings within `helpers.ts` because we're going multichain. I walked through with butler and agreed that domain specific makes more sense in this case.

@butlerji had thoughts to bundle his PR on updating deployment in here too. This may/may not happen in time though...

protocol/contracts/governance/Templar.sol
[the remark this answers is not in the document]
frontier159: If normalising to optimise gas/space was indeed critical (which I don't think it is given the number of discordIds we'll need):
Could also have a dynamic map linking an `roleIndex->roleString`, so then the value of templeRole is also a uint256.

protocol/contracts/governance/Templar.sol
nichosystem: nope doesn't make client side harder

protocol/contracts/governance/ElderElection.sol
nichosystem: definitely would make it easier to drop the nonces from the signature -- I don't see an effective replay attack here

protocol/contracts/governance/Templar.sol
mountainpath9: > We don't actually care if the token was transferred or mint

We might eventually. It could well be useful to know when a discord user first went on chain (ie mint), and when a discord user changed their address (ie transfer)?

mountainpath9: I don't have experience with how much flexibility thegraph supports, but I assume it would be straightforward to produce 2 subgraphs

- one mapping address to discord id   (using Mint and Transfer events)
- the other mapping discord id to temple role (using the UpdateTempleRole event)

the election results calculated could use these as inputs directly.

protocol/contracts/governance/Templar.sol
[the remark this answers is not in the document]
mountainpath9: I'd be open to that if we need it, but let's keep things simple for now.

protocol/contracts/governance/Templar.sol
mountainpath9: Updating the roles and assigning the NFTS are now seperate methods on 2 different contracts. So I'm, resolving this, as we can start a new thread if required.

protocol/contracts/governance/ElderElection.sol
mountainpath9: Would this be a possible replay attack:

   - User creates a signed TX1 to set endorsements via the relay
   - User changes endorsements be creating a signed TX2 and sending to the relay
   - Hacker undoes TX2 by replaying TX1

protocol/contracts/governance/TemplarMetadata.sol
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * Set the (highest) temple role held by a templar with the given discordid
+     */
+    function setRole( 
+        uint256 discordId,
+        string calldata _templeRole
+    ) external onlyRole(CAN_UPDATE) {
+        templars.checkExists(discordId);
```
frontier159, line 32: I wonder if this should just return a bool such that the client can decide if it will revert or not

frontier159 (approved):

protocol/contracts/governance/TemplarMetadata.sol
mountainpath9: It was a conscious decision so that there was a standard error that was thrown. 

But an addition function that returns a bool would be useful - will add.

protocol/contracts/governance/ElderElection.sol
nichosystem: that's true, if they undo their endorsement before the deadline then the first tx could be replayed

nichosystem (approved):

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,142 @@
...
+     */
+    function _useNonce(address _owner) internal returns (uint256 current) {
+        Counters.Counter storage nonce = nonces[_owner];
+        current = nonce.current();
+        nonce.increment();
+    }
+
+    event UpdateNomination(uint256 indexed discordId, bool isNominated);
+    event UpdateEndorsements(address indexed account, uint256[] discordId);
+    event UpdateNominationFee(uint256 nominationFee);
```
medariox, line 132: Leftover event from nomination fee - can be removed.

protocol/contracts/governance/ElderElection.sol
```diff
@@ -0,0 +1,141 @@
...
+     * "Consume a nonce": return the current value and increment.
+     */
+    function _useNonce(address _owner) internal returns (uint256 current) {
+        Counters.Counter storage nonce = nonces[_owner];
+        current = nonce.current();
+        nonce.increment();
+    }
+
+    event UpdateNomination(uint256 indexed discordId, bool isNominated);
+    event UpdateEndorsements(address indexed account, uint256[] discordId);
```
medariox, line 131: Very minor, but this should be uint256[] discordId**s**.

protocol/contracts/governance/TemplarMetadata.sol
mountainpath9: This is done.

PR #620, opened 2022-11-02
feat: update zaps for approved targets

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/zaps/TempleZaps.sol
```diff
@@ -232,7 +232,8 @@ contract TempleZaps is ZapBase {
     address token1 = IUniswapV2Pair(pair).token1();
 
     if (_fromAddress != token0 && _fromAddress != token1) {
-
+      require(approvedTargets[_fromAddress][_swapTarget], "TempleZaps: Unsupported token/target");
```
frontier159, line 235: @princetonbishop It's a little odd that the `approvedTargets` are on this TempleZaps (via the ZapBase inheritance) too, but currently unused.

What were you thinking here - that TempleZaps has it's own allow list, or that it shares the same list as GenericZap (in which case we'd change this line to use `zaps.approvedTargets()`?

protocol/contracts/zaps/GenericZap.sol
```diff
@@ -192,7 +192,7 @@ contract GenericZap is ZapBase {
     address swapTarget,
```
frontier159, line 192: At the top of this contract -- In order for this to work in testnet, would have to change:
```
  ICurveFactory private immutable curveFactory = ICurveFactory(0xB9fC157394Af804a3578134A6585C0dc9cc990d4);
  IBalancerVault private immutable balancerVault = IBalancerVault(0xBA12222222228d8Ba445958a75a0704d566BF2C8);
```
To be set in the constructor too...know what the goerli addresses are (if any)?

protocol/contracts/zaps/TempleZaps.sol
frontier159: Also, I can't see any tests which check that the unsupported token/target fails...might be worth setting up solidity coverage in temple core?

protocol/contracts/zaps/GenericZap.sol
princetonbishop: `balancerVault` can be found here
https://goerli.etherscan.io/address/0xBA12222222228d8Ba445958a75a0704d566BF2C8

protocol/contracts/zaps/TempleZaps.sol
princetonbishop: for consistency across zaps, making `templeZaps` use zaps approved targets is more ideal.

frontier159: Works for me -- should probably remove from ZapBase and put directly into GenericZaps then I think

protocol/contracts/zaps/GenericZap.sol
frontier159: Will 0x on goerli have issues routing to the right pools if they're self-created?

Issue #459, opened 2022-07-21
Fix Core Notifications


mrfujisawa:
The Notification Provider in Core is a mess.  Generally speaking, the code and UI could be cleaned up.

Two main places for improvement:
1. The notifications are absolutely positioned and should be fixed.  If you've scrolled down the page, you wont see a notification unless you scroll up
2. The notification manager should manage showing/animating the notification itself.  Right now the Notification has a lot of complex logic and knows how to animate itself/it could be simplified.

MrMysteryMeat:
currently working on this

PR #644, opened 2022-11-23
Fix core notifications

MrMysteryMeat:
# Description

As Fuji described in https://github.com/TempleDAO/temple/issues/459, notifications should be fixed, so that you don't have to scroll up to see them, and the notification manager should handle closing notifications. I removed a lot of the complex logic in Notifications.tsx and changed it so that NotificationManager.tsx handles the timing of notifications. The code should be much more clear, I also added comments.

Fixes # (issue)

# Checklist
- [X] Code follows the style guide
- [X] I have performed a self-review of my own code
- [X] New and existing tests pass locally
- [X] This PR is targeting the correct branch


MrMysteryMeat:
BY THE WAY: in staging, all notifications opened within the ten second time frame of the first notification will close once the timeframe of the first notification ends (with the same hash). Meaning that if the user gets a notification and then gets a notification nine seconds later, the second notification will only show for one second. In this version, all notifications will close 10 seconds after the last notification thrown, regardless of the hash

apps/dapp/src/components/Notification/NotificationManager.tsx
```diff
@@ -1,16 +1,54 @@
...
+
+const AUTOHIDE_TIMER = 1000 * 10;
 
 const NotificationManager = () => {
   const { notifications } = useNotification();
 
+  const { closeNotification } = useNotification();
+
+  // Array indicating which notifications are open.
+  // I know what you're thinking - "couldn't this just
```
nichosystem, line 14: appreciate the context, we can cut this extended comment for now 🙂

apps/dapp/src/components/Notification/NotificationManager.tsx
```diff
@@ -1,16 +1,54 @@
...
+  // be a boolean because we are always setting the whole
+  // array to either true of false?" Well I tested that,
+  // and it messes up the animation when opening multiple
+  // notifications at once, so let's keep it like this.
+  const [areOpen, setAreOpen] = useState<boolean[]>([]);
+
+  useEffect(() => {
+    // Setting all notifications to be open by default
+    var listTrue = new Array(notifications.length).fill(true);
```
nichosystem, line 23: nit: let's stick to lets or consts for consistency

nichosystem (changes requested):
just two nits to clean this up otherwise it looks good 👍

apps/dapp/src/components/Notification/NotificationManager.tsx
MrMysteryMeat: done

apps/dapp/src/components/Notification/NotificationManager.tsx
MrMysteryMeat: done

nichosystem (approved):

Issue #447, opened 2022-07-13
Automate etherscan verify as a githook and add as a merge blocking condition


butlerji:
This is an important step to allow  Decentralised Contract Deployment, as we have separated deployment from verification, and can be sure what's checked into git is what's deployed on mainnet/testnet.

BoxedFruits:
I am interested in picking this up but want to make sure that I am understanding the ticket correctly.

Whenever there is a new deploy of contracts [helper.ts](https://github.com/TempleDAO/temple/blob/stage/protocol/scripts/deploys/helpers.ts#L52) _should_ be updated. What we want to add is that if the constants mentioned [here](https://github.com/TempleDAO/temple/tree/stage/protocol/scripts/deploys/mainnet/core-deploy) are changed because of a new deployement, there should be a githook that blocks merging to the main branch if it is not verified?

PR #656, opened 2022-12-06
Auto Verify Contract

MrMysteryMeat:
# Description
This PR serves to automate etherscan verify after deploy in helper.ts. It relies on whether or not process.env.ETHERSCAN_API_KEY exists, which I don't have locally but I'm sure it exists in prod no? That being said I have yet to test this code. though I am pretty confident it will work. I am going to look into how to test this, but any PR reviewers that see this and have time and want to test this are welcome to (in fact I would appreciate it). Just throwing a PR down while I dig into how to test this.

Fixes # (issue)

https://github.com/TempleDAO/temple/issues/447

12/14/2022 UPDATE: I have gotten to a point to where I have absolutely no idea why this is not working. The API call seems to be correct, and it even returns a success message; however when I check the transaction on goerli.etherscan is does not say that the contract is actually verified. All of the parameters for the API call seem to be correct, I even used this flattener https://github.com/EricxStone/sol-flattener to flatten the .sol files sourcecode with imports.

The way that the flattener gets the path for the .sol files is a bit hacky, but it works.

All of the pieces of the puzzle are here I just cant actually get a contract to verify without manually typing the command that DeployAndMine spits out.

Part of me believes that the API itself is simply broken. plz halp

# Checklist
- [X] Code follows the style guide
- [X] I have performed a self-review of my own code
- [?] New and existing tests pass locally
- [X] This PR is targeting the correct branch


MrMysteryMeat:
Protocol Tests / Hardhat Tests (pull_request) fails even if the auto verification code is not added... not sure whats going on

MrMysteryMeat:
NOTE: currently set to run on only hardhat chain we can change that easily

protocol/scripts/deploys/helpers.ts
```diff
@@ -269,6 +270,32 @@ export async function deployAndMine<
...
+  // Wait for a few blocks to be mined before verifying
+  await contract.deployTransaction.wait(6)
+  // Verify the contract on Etherscan
+  const response = await axios.post(
+    `https://api.etherscan.io/api?module=contract&action=verifysourcecode&apikey=${process.env.ETHERSCAN_API_KEY}`,
+    {
+      contractaddress: contract.address,
+      sourceCode: (contract as any).deployedBytecode,
+      contractname: name,
+      compilerversion: 'v0.7.0+commit.9e61f92b',
```
nichosystem, line 284: most of our stuff is v0.8.0 but this can vary. You should be able to pull it from the hardhat config I believe?

nichosystem (changes requested):
very cool. definitely worth testing, get some goerli eth and run a testnet deployAndMine script, see if it succeeds

protocol/scripts/deploys/helpers.ts
```diff
@@ -269,6 +270,32 @@ export async function deployAndMine<
...
+      optimizationUsed: 1,
+      runs: 200,
+      constructorArguements: renderedArgs,
```
frontier159, line 287: Typo in arguements

Also need to check this escapes multi word string args correctly

frontier159: And also for arrays

protocol/scripts/deploys/helpers.ts
MrMysteryMeat: gonna get crackin on this td

MrMysteryMeat:
UPDATE: I have the endpoint working in postman, i just need a way to get the source code for each contract that is to be deployed programmatically. This is what i'm currently looking into, any help would be greatly appreciated <3

PR #746, opened 2023-01-30
Deploy Epoch 13 team payments

nichosystem:
# Description
Updated the team payments page so all we have to do is add a new address to the env for the page to update

# Checklist


apps/dapp/src/constants/env/production.tsx
```diff
@@ -20,18 +20,21 @@ const env: Environment = {
...
-    teamPaymentsEpoch4: '0x07888e0a8929eb922Aee5930f7B0894BaB5D8120',
-    teamPaymentsEpoch5: '0xc52101f2e278a316c70cc233486bce1863606920',
-    teamPaymentsEpoch6: '0x91ad65e053ae98b4fbab84fc38b7bddd17c32cda',
-    teamPaymentsEpoch7: '0x8cded928006feb238617fa5f7b04abeefcde36bf',
-    teamPaymentsEpoch8: '0x008eCB3E53024628a5A8BbE0b142329791ad6f51',
-    teamPaymentsEpoch9: '0x5461d7Cd3eEB184a83c5a1678335D72ccaf04818',
-    teamPaymentsEpoch10: '0xdace5aa4D8E1E2678a2F2F6E96F31eBD599503DB',
-    teamPaymentsEpoch11: '0x49fb6dbe198f61d8962cb069ca1bc7f2daff4de6',
-    teamPaymentsEpoch12: '0x465451535c4518d805cbead0b95e1a1a677ddeae',
+    teamPayments: [
```
marshall2112, line 23: Considering the order is important, I'd add a warning/comment here in the code to instruct others to preserve the order. I guess the only side effect would be the Epoch number would show up wrong in the UI but still would be confusing.

marshall2112 (changes requested):

marshall2112 (approved):

PR #788, opened 2023-04-07
feat: Temple v2 & TLC

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/interfaces/v2/ITempleStrategyManager.sol
```diff
@@ -0,0 +1,120 @@
...
+
+        /// @dev The total amount of dUSD the strategy can borrow.
+        /// Note: The strategy may end up accruing more debt than this ceiling as it accrues
+        uint256 debtCeiling;
+
+        /// @dev When the strategies totalEquity() is below this amount then operators
+        /// should consider unwinding positions and paying down debt.
+        /// A soft limit for reporting purposes only.
+        uint256 underperformingEquityThreshold;
+    }
```
temple-advocate, line 30: /// @dev Governance nominates this strategy to auto compound into underlying position at select positive Equity threshold.
bool autoCompounding;

/// @ dev when the strategies totalEquity() is above this amount then operators should sell accrued yield and compound into underlying position.
uin256 autocompoundingThreshold;

frontier159: These would make sense for specific strategies - but not all, correct?

I think these parameters could be added just to those strategies where it makes sense (ie each strategy has an individual set of params which could be modified)

temple-advocate: So you think that each contract should be bespoke or use existing parameters as much as possible?

frontier159: They'll be bespoke parameters for each strategy. It gives flexibility later for the strategy to do what it needs to, and means we won't have to redeploy the manager each time.

It also means the won't be a bunch of non applicable params or params used for something different than what the name of it suggests

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
```diff
@@ -0,0 +1,428 @@
...
+        // EXACT_BPT_IN_FOR_TOKENS_OUT index is 1 for exitKind
+        bytes memory encodedUserdata = abi.encode(uint256(1), bptAmount);
+        requestData.userData = encodedUserdata;
+        requestData.toInternalBalance = false;
+
+        (, requestData.minAmountsOut) = balancerHelpers.queryExit(balancerPoolId, amo, amo, requestData);
+        (expectedProtocolTokenAmount, expectedQuoteTokensAmount) = protocolTokenIndexInBalancerPool == 0
+            ? (requestData.minAmountsOut[0], requestData.minAmountsOut[1])
+            : (requestData.minAmountsOut[1], requestData.minAmountsOut[0]);
+        minProtocolTokenAmount = applySlippage(expectedProtocolTokenAmount, slippageBps);
```
princetonbishop, line 409: nit: have in mind if slippage is large enough, could lead to frontrunning. perhaps a way to control with maxSlippage?

protocol/contracts/amo/Ramos.sol
```diff
@@ -191,207 +264,304 @@ contract RAMOS is Ownable, Pausable {
...
-        stable.safeTransfer(address(poolHelper), amountIn);
-        // stable join
-        uint256 joinTokenIndex = templeBalancerPoolIndex == 0 ? 1 : 0;
-        uint256 bptOut = poolHelper.joinPool(
-            amountIn, minBptOut, rebalancePercentageBoundUp, rebalancePercentageBoundLow,
-            templePriceFloorNumerator, postRebalanceSlippage, joinTokenIndex, stable
-        );
+        // Collect the fees from the input quote token
+        uint256 feeAmt = quoteTokenAmountIn * rebalanceFees.rebalanceJoinFeeBps / BPS_PRECISION;
+        if (feeAmt != 0) {
```
princetonbishop, line 371: nit: I would add FeeCollected event

protocol/contracts/amo/Ramos.sol
```diff
@@ -191,207 +264,304 @@ contract RAMOS is Ownable, Pausable {
...
      * BPT tokens are then deposited and staked in Aura.
      * @param request Request data for joining balancer pool. Assumes userdata of request is
      * encoded with EXACT_TOKENS_IN_FOR_BPT_OUT type
-     * @param minBptOut Minimum amount of BPT tokens expected to receive
      */
     function addLiquidity(
-        AMO__IBalancerVault.JoinPoolRequest memory request,
-        uint256 minBptOut
-    ) external onlyOwner {
+        IBalancerVault.JoinPoolRequest memory request
```
princetonbishop, line 458: nit: gas savings: we could use `calldata` here as request is not modified

protocol/contracts/amo/Ramos.sol
```diff
@@ -191,207 +264,304 @@ contract RAMOS is Ownable, Pausable {
...
+     * @param to Address to which the `quoteToken` received from balancer pool are transferred
      */
     function removeLiquidity(
-        AMO__IBalancerVault.ExitPoolRequest memory request,
-        uint256 bptIn
-    ) external onlyOwner {
+        IBalancerVault.ExitPoolRequest memory request,
+        uint256 bptIn,
+        address to
+    ) external override onlyElevatedAccess {
```
princetonbishop, line 529: RAMOS has a number of guarded functions. perhaps we could use `payable` to save some gas.
compiler will ignore checks for if payment was provided.

protocol/contracts/v2/access/TempleElevatedAccess.sol
```diff
@@ -0,0 +1,107 @@
...
+    function setRescueMode(bool value) external override {
+        if (!rescuers[msg.sender]) revert CommonEventsAndErrors.InvalidAccess();
+        emit RescueModeSet(value);
+        inRescueMode = value;
+    }
+
+    /**
+     * @notice Grant `account` the emergency operations role
+     */
+    function setRescuer(address account, bool value) external override onlyElevatedAccess {
```
princetonbishop, line 52: I would use the set, accept format for these critical functions/roles. eg. like FRAX does `setAdmin`, `acceptAdmin`

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
```diff
@@ -0,0 +1,428 @@
...
+import { IBalancerPoolHelper } from "contracts/interfaces/amo/helpers/IBalancerPoolHelper.sol";
+import { IBalancerVault } from "contracts/interfaces/external/balancer/IBalancerVault.sol";
+import { IBalancerHelpers } from "contracts/interfaces/external/balancer/IBalancerHelpers.sol";
+import { AMOCommon } from "contracts/amo/helpers/AMOCommon.sol";
+
+interface IWeightPool2Tokens {
+    function getNormalizedWeights() external view returns (uint256[] memory);
+}
+
+contract BalancerPoolHelper is IBalancerPoolHelper {
```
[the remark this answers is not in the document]
frontier159: Perhaps - but maybe just at the Ramos level rather than on the pool helper? The `protocolToken` will give it away too imo

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
```diff
@@ -0,0 +1,428 @@
...
+    IERC20 public immutable override bptToken;
+    IERC20 public immutable override protocolToken;
+    IERC20 public immutable override quoteToken;
+    address public immutable override amo;
+    
+    uint256 public constant override BPS_PRECISION = 10_000;
+    uint256 public constant override PRICE_PRECISION = 1e18;
+
+    // @notice protocolToken index in balancer pool
+    uint64 public immutable override protocolTokenIndexInBalancerPool;
```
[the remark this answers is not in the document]
frontier159: I just learnt something new. For some reason, I had it in my head that byte packing only happened to structs, rather than on top level primitives too.

You're right!

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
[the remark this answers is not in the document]
frontier159: My view is that it's a concern of the bot though, not of the contract. Eg uniswap/curve/bal doesn't limit what you can enter here.

frontier159: same

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
```diff
@@ -0,0 +1,428 @@
...
+        minQuoteTokensAmount = applySlippage(expectedQuoteTokensAmount, slippageBps);
+
+        // update `requestData` with the `minAmountsOut` to which the slippage was applied
+        (requestData.minAmountsOut[0], requestData.minAmountsOut[1]) = protocolTokenIndexInBalancerPool == 0
+            ? (minProtocolTokenAmount, minQuoteTokensAmount)
+            : (minQuoteTokensAmount, minProtocolTokenAmount);
+    }
+
+    function applySlippage(uint256 amountIn, uint256 slippageBps) public override pure returns (uint256 amountOut) {
+        amountOut = amountIn * (BPS_PRECISION - slippageBps) / BPS_PRECISION;
```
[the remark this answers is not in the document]
frontier159: The calc will fail anyway if it's `> BPS_PRECISION` since that'll be a negative number. So don't think it's worth the extra check

protocol/contracts/amo/Ramos.sol
```diff
@@ -119,53 +156,89 @@ contract RAMOS is Ownable, Pausable {
...
+        emit TpiOracleSet(newTpiOracle);
+        tpiOracle = ITreasuryPriceIndexOracle(newTpiOracle);
+    }
 
-        emit SetTemplePriceFloorNumerator(_numerator);
+    /**
+     * @notice Set the Treasury Price Index (TPI) Oracle
+     */
+    function setProtocolTokenVault(address vault) external override onlyElevatedAccess {
+        emit ProtocolTokenVaultSet(vault);
```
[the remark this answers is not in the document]
frontier159: Slither liberally warns about 'reentrancy' issues - where events may not be fired if external contracts are called prior to the emit, then re-entered. So I got into the habit of emitting events as early as possible.

protocol/contracts/amo/Ramos.sol
```diff
@@ -119,53 +156,89 @@ contract RAMOS is Ownable, Pausable {
...
+
```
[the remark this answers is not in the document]
frontier159: Is this extra check warranted for the bytecode increase? It will only be called rarely by elevated access, and it wouldn't be the case that the new vault is the same as the old one

protocol/contracts/amo/Ramos.sol
```diff
@@ -191,207 +264,304 @@ contract RAMOS is Ownable, Pausable {
      * @param to Recipient address
      * @param amount Amount to recover
      */
-    function recoverToken(address token, address to, uint256 amount) external onlyOwner {
+    function recoverToken(address token, address to, uint256 amount) external onlyElevatedAccess {
```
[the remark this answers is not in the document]
frontier159: To confirm, are you suggesting that we always send the funds to the 'executor' or 'rescuer' role, rather than as a `to` param?

In the case that someone accidentally transferred a token to this contract and want it back, having the `to` allows us to send it straight back to the user, rather than multiple tx's.

`recoverToken` would be an example for a higher threshold of signers (eg 3 or 4)

protocol/contracts/amo/Ramos.sol
frontier159: Great idea

protocol/contracts/amo/Ramos.sol
frontier159: This is a lvl11 gas optimisation -- I'm not sure I follow this one sorry. What checks are ignored, how much gas andis it worth it for the extra confusion (for us, auditors, partners using RAMOS) since we'll only be accepting ERC20's?

protocol/contracts/amo/Ramos.sol
frontier159: Hmm - there's a really strange bytesize (ie `yarn size`) issue if I change this to calldata. 

as `memory`:   ` |  Ramos                            ·      23.776  ·                |`
as `calldata`: ` |  Ramos                            ·      24.541  ·        +0.765  │`

So we're right up close to the size limit as it is, and then calldata blows past it. So while it's a small gas saving, I don't think we can use it.

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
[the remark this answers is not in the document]
princetonbishop: now that you mention, RAMOS level also makes more sense

protocol/contracts/amo/Ramos.sol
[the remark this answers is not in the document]
princetonbishop: no, I wasn't suggesting we send to the `executor/rescuer`, i meant a predefined mutable state variable (destination ) that could be set be elevatedAccess.
Considering the latter case of user locked funds, having it unchanged skips an extra transfer. So you can leave it as is

protocol/contracts/amo/Ramos.sol
princetonbishop: ah I see, now that reminds me why i left it initially as memory, my bad. thanks!

protocol/contracts/amo/Ramos.sol
princetonbishop: I think a lot of auditors are aware of this gas optimization. But if you think this creates confusion in the long run for others, then ignore.
Here's an article with some explanation 
https://dev.to/zaryab2000/does-adding-a-payable-keyword-in-solidity-actually-save-gas-24bm

frontier159: So just 24 gas saving? Seems a bit over the top imo

protocol/contracts/v2/access/TempleElevatedAccess.sol
frontier159: Ah yes, nice one will do

protocol/contracts/amo/Ramos.sol
frontier159: It's a bizarre side effect. Gheeze the solidity compiler could do with some work!

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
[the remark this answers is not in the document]
frontier159: I'm not sold on needing this tbh, given the combination of `[protocolToken, quoteToken]` already tells us all we need.

Also I can't add any more params to the constructor or we get a stack too deep. So would need to refactor to take a struct instead...all means more bytecode

protocol/contracts/amo/Ramos.sol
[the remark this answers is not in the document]
frontier159: Decided not to implement

protocol/contracts/amo/Ramos.sol
frontier159: Done in https://github.com/TempleDAO/temple/pull/814/files

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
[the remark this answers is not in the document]
frontier159: Decided not to implement

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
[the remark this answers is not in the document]
frontier159: Done in https://github.com/TempleDAO/temple/pull/814/files

protocol/contracts/amo/helpers/BalancerPoolHelper.sol
frontier159: Decided not to implement - responsibility of the bot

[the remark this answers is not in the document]
frontier159: Decided not to implement - responsibility of the bot

protocol/contracts/amo/Ramos.sol
```diff
@@ -119,53 +156,89 @@ contract RAMOS is Ownable, Pausable {
...
+     * @notice Set the Treasury Price Index (TPI) Oracle
+     */
+    function setTpiOracle(address newTpiOracle) external override onlyElevatedAccess {
+        emit TpiOracleSet(newTpiOracle);
+        tpiOracle = ITreasuryPriceIndexOracle(newTpiOracle);
+    }
 
-        emit SetTemplePriceFloorNumerator(_numerator);
+    /**
+     * @notice Set the Treasury Price Index (TPI) Oracle
```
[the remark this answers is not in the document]
frontier159: Done in https://github.com/TempleDAO/temple/pull/814/files

protocol/contracts/amo/Ramos.sol
frontier159: Decided not to implement

protocol/contracts/v2/access/TempleElevatedAccess.sol
frontier159: Done: https://github.com/TempleDAO/temple/pull/814/commits/12985afec975531fc88e804859d7d6565742f903

protocol/contracts/v2/TreasuryReservesVault.sol
```diff
@@ -0,0 +1,474 @@
...
+    function repayTemple(uint256 repayAmount, address strategy) external override {
+        if (repayAmount == 0) revert CommonEventsAndErrors.ExpectedNonZero();
+        if (globalRepaysPaused) revert RepaysPaused();
+
+        Strategy storage strategyData = strategies[strategy];
+        if (!strategyData.isEnabled) revert NotEnabled();
+        if (strategyData.repaysPaused) revert RepaysPaused();
+
+        // dUSD price = Temple * TPI
+        uint256 debtToBurn = repayAmount * treasuryPriceIndex() / 1e18;
```
princetonbishop, line 364: I would add a safety check to check if tpi is not 0 or invalid

protocol/contracts/v2/TreasuryReservesVault.sol
```diff
@@ -0,0 +1,474 @@
...
+
+        // @todo need an event like:?
```
princetonbishop, line 366: could also add token address as argument to `Repay` event to differentiate tokens

protocol/contracts/v2/TreasuryReservesVault.sol
```diff
@@ -0,0 +1,474 @@
...
+     * The strategy executor is responsible for unwinding all it's positions first and sending stables to the TRV.
+     * All outstanding dUSD debt is burned, leaving a net gain/loss of equity for the shutdown strategy.
+     */
+    function shutdown(address strategyAddr) external override {
+        if (msg.sender != strategyAddr && !isElevatedAccess(msg.sender, msg.sig)) revert CommonEventsAndErrors.InvalidAccess();
+
+        Strategy storage strategyData = strategies[strategyAddr];
+        if (!strategyData.isEnabled) revert NotEnabled();
+        if (!strategyData.isShuttingDown) revert NotShuttingDown();
+
```
princetonbishop, line 448: by design, perhaps add a way to ensure strategy is unwinded before shutting down (in the event `automatedShutdown` isn't called)?
This could lead to a Denial of service (for Strategy) or possible inaccurate calculation (dashboard) if executors mistakenly execute `setShuttingDown` and `shutdown` immediately one after the next

protocol/contracts/v2/strategies/DsrBaseStrategy.sol
```diff
@@ -0,0 +1,287 @@
...
+     * This may be stale if the DSR rate hasn't been updated for a while.
+     */
+    function latestDsrBalance() public view returns (uint256) {
+        // pot.chi() == DAI/share (accurate to the last checkpoint)
+        // pot.pie() == number of dsr shares this strategy holds.
+        // pot.rho() == Timestamp of last checkpoint
+
+        uint256 rho = pot.rho();
+        // solhint-disable-next-line not-rely-on-time
+        uint256 chi = (block.timestamp > rho) ? _rpow(pot.dsr(), block.timestamp - rho) * pot.chi() / RAY : pot.chi();
```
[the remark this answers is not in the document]
frontier159: `pot.drip()` does do the `*chi/1e27`

see the rmul here:
```
        tmp = rmul(rpow(dsr, now - rho, ONE), chi);
```

FWIW - this is the same as sDAI -- Maker's new Aave copy pasta (https://vscode.blockscan.com/ethereum/0x83f20f44975d03b1b09e64809b757c47f942beea)

sDAI a vault token which just wraps DSR

protocol/contracts/v2/interestRate/BaseInterestRateModel.sol
```diff
@@ -0,0 +1,41 @@
...
+        // Cap the UR at 100%
+        if (utilizationRatio > PRECISION) {
+            utilizationRatio = PRECISION;
+        }
+
+        interestRate = computeInterestRateImpl(utilizationRatio);
+
+        if (interestRate > MAX_ALLOWED_INTEREST_RATE) {
+            interestRate = MAX_ALLOWED_INTEREST_RATE;
+        } else if (interestRate < MIN_ALLOWED_INTEREST_RATE) {
```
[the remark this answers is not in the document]
frontier159: Good catch. 

Fixed in https://github.com/TempleDAO/temple/pull/816/commits/cbdd48eaa931ba8cb9407c55ca41ffc53a831d91

protocol/contracts/v2/TempleDebtToken.sol
```diff
@@ -0,0 +1,583 @@
...
+            // Update the per debtor checkpoint of (risk premium) interest, and also update the 
+            // cumulative estimate of total debtor interest owing.
+            unchecked {
+                debtor.checkpoint = uint160(_debtorCheckpoint - _debtorDebtRepaid);
+                estimatedTotalRiskPremiumInterest -= _debtorDebtRepaid;
+            }
+        }
+    }
+
+    function _burnDebtorInterest(
```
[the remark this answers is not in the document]
frontier159: agreed - will fix

protocol/contracts/v2/TreasuryReservesVault.sol
```diff
@@ -0,0 +1,474 @@
...
+        if (!strategyData.isEnabled) revert NotEnabled();
+        emit StrategyPausedSet(_strategy, _pauseBorrow, _pauseRepays);
+        strategyData.borrowPaused = _pauseBorrow;
+        strategyData.repaysPaused = _pauseRepays;
+    }
+
+    /**
+     * Add a new strategy
+     */
+    function addNewStrategy(address strategy, uint256 debtCeiling, int256 underperformingEquityThreshold) external override onlyElevatedAccess {
```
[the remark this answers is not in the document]
frontier159: Yep that's right - given all the data is deleted, it's fine to be added again later (and not worth adding a guard)

protocol/contracts/v2/TreasuryReservesVault.sol
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /**
+     * @notice Update the underperforming equity threshold.
+     */
+    function setStrategyUnderperformingThreshold(address strategy, int256 underperformingEquityThreshold) external override onlyElevatedAccess {
+        Strategy storage strategyData = strategies[strategy];
+        if (!strategyData.isEnabled) revert NotEnabled();
+        if (underperformingEquityThreshold == 0) revert CommonEventsAndErrors.ExpectedNonZero();
+        emit UnderperformingEquityThresholdUpdated(strategy, strategyData.underperformingEquityThreshold, underperformingEquityThreshold);
```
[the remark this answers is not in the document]
frontier159: Given this is in an admin-only and not called frequently, I don' think it's worth adding

protocol/contracts/v2/TreasuryReservesVault.sol
frontier159: There are checks on the `TreasuryPriceIndexOracle::setTreasuryPriceIndex()`, when it's set. So don't think think it's worth adding here too.

There's a 'max delta' to prevent a fat finger update of the TPI

protocol/contracts/v2/TreasuryReservesVault.sol
frontier159: Yes...so there was a change in spec, as is the temple way! I didn't want to pollute your current PR though

I'm currently rejigging the TRV such that it can be configured to borrow/repay any configured token (and debt token pair).

I'll give you details in Discord.

protocol/contracts/v2/TreasuryReservesVault.sol
frontier159: I don't think we can it 'by design'. For example if it's a Gnosis strategy, there's no easy way we can guarantee it's been fully unwound.

However I think this is going to be reasonably resilient:
1. The two steps means the TRV has to initiate it first
2. The shutdown() method will be more locked down - it will need a higher threshold of signers to call. So the idea is that it's thoroughly checked

Regardless of the shutdown, the AbstractStrategy still has a recoverToken() too, so can be pulled out after the fact

protocol/contracts/v2/strategies/GnosisStrategy.sol
```diff
@@ -0,0 +1,149 @@
...
+    event Repay(uint256 amount);
+
+    constructor(
+        address _initialRescuer,
+        address _initialExecutor,
+        string memory _strategyName,
+        address _treasuryReservesVault,
+        address _gnosisSafeWallet
+    ) AbstractStrategy(_initialRescuer, _initialExecutor, _strategyName, _treasuryReservesVault) {
+        gnosisSafeWallet = _gnosisSafeWallet;
```
princetonbishop, line 35: I see no setters, make immutable?

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        (shutdownParams.bptAmount,,) = ramos.positions();
+        (,,,,shutdownParams.requestData) = proportionalRemoveLiquidityQuote(shutdownParams.bptAmount, populateParams.slippageBps);
+        shutdownData = abi.encode(shutdownParams);
+    }
+
+    /**
+     * @notice Shutdown the strategy.
+     * First unstake all BPT and liquidate into stables, and then repay the stables to the TRV.
+     * @param shutdownData abi encoded data of struct `PopulateShutdownParams`
```
princetonbishop, line 175: fix up comment

protocol/contracts/v2/TreasuryReservesVault.sol
princetonbishop: Yeah I realized this after I reviewed the GnosisStrategy. 
Okay got it!

protocol/contracts/v2/TreasuryReservesVault.sol
princetonbishop: learned a new word, rejigging lol

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,968 @@
...
+
+        // burn the temple collateral by repaying to TRV. This will burn the equivalent dUSD debt too.
+        if (totalCollateralClaimed != 0) {
+            templeToken.safeIncreaseAllowance(address(treasuryReservesVault), totalCollateralClaimed);
+            treasuryReservesVault.repayTemple(totalCollateralClaimed, address(tlcStrategy));
+            totalCollateral -= totalCollateralClaimed;
+        }
+
+        // Remove debt from the totals
+        repayTotalDebt(_debtTokenCache, totalDebtWiped.encodeUInt128());
```
princetonbishop, line 389: `if totalDebtWiped > 0`

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,968 @@
...
+     * @notice Set the Withdrawal Collateral Request window parameters
+     * @param minSecs The number of seconds which must elapse between a request and the action
+     * @param maxSecs The number of seconds until a request expires
+     */
+    function setWithdrawCollateralRequestConfig(
+        uint256 minSecs,
+        uint256 maxSecs
+    ) external override onlyElevatedAccess {
+        emit RemoveCollateralRequestConfigSet(minSecs, maxSecs);
+        removeCollateralRequestConfig = FundsRequestConfig(uint32(minSecs), uint32(maxSecs));
```
princetonbishop, line 436: revert if `minSecs >= maxSecs`

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,968 @@
...
+    /**
+     * @notice Set the Borrow Request window parameters
+     * @param minSecs The number of seconds which must elapse between a request and the action
+     * @param maxSecs The number of seconds until a request expires
+     */
+    function setBorrowRequestConfig(
+        uint256 minSecs,
+        uint256 maxSecs
+    ) external override onlyElevatedAccess {
+        emit BorrowRequestConfigSet(minSecs, maxSecs);
```
princetonbishop, line 448: revert if `minSecs >= maxSecs`

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,968 @@
...
+        DebtTokenCache memory _debtTokenCache,
+        uint256 _collateral,
+        uint256 _debt
+    ) internal pure returns (uint256) {
+        return _debt == 0
+            ? type(uint256).max
+            : mulDiv(
+                _collateral * _debtTokenCache.price,
+                _debtTokenCache.config.maxLtvRatio,
+                _debt * 1e18
```
princetonbishop, line 946: nit: you could define 1e36 and 1e18 as constants

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
```diff
@@ -0,0 +1,236 @@
...
+        emit SafeTxExecutorRemoved(executor);
+        if (!_safeTxExecutors.remove(executor)) revert InvalidExecutor();
+    }
+
+    /**
+      * @notice Set the default number of signatories required, when a contract/function signature
+      * not explicitly defined.
+     */
+    function setDefaultSignaturesThreshold(uint256 threshold) external onlyElevatedAccess {
+        emit DefaultSignaturesThresholdSet(threshold);
```
princetonbishop, line 97: can we check just to be sure this is not set to 0

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
```diff
@@ -0,0 +1,236 @@
...
+        defaultSignaturesThreshold = threshold;
+    }
+
+    /**
+      * @notice Set the number of signatories required for a contract/function signature pair.
+     */
+    function setFunctionThreshold(address contractAddr, bytes4 functionSignature, uint256 threshold) external onlyElevatedAccess {
+        if (contractAddr == address(0)) revert InvalidAddress();
+        if (functionSignature == bytes4(0)) revert InvalidFunctionSignature();
+
```
princetonbishop, line 107: nit:
```
 assembly { 
   if not(gt(extcodesize(contractAddr), 0)) {
      revert(mload(MSG_LOCATION), MSG_SIZE)
   }
 }
```

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
```diff
@@ -0,0 +1,236 @@
...
+      * @notice Set the number of signatories required for a number of function signatures at a time.
+     */
+    function setFunctionThresholdBatch(address contractAddr, bytes4[] memory functionSignatures, uint256 threshold) external onlyElevatedAccess {
+        if (contractAddr == address(0)) revert InvalidAddress();
+
+        uint256 length = functionSignatures.length;
+        bytes4 sig;
+        for (uint256 i; i < length; ++i) {
+            sig = functionSignatures[i];
+            if (sig == bytes4(0)) revert InvalidFunctionSignature();
```
princetonbishop, line 122: don't you rather want to skip instead of reverting?

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
```diff
@@ -0,0 +1,236 @@
...
+
+            // When Tenderly simulations run, the Safe's threshold is overriden to 1.
+            // So only check the signatures when the threshold is > 1
+            // WARNING: This means that we aren't doing any extra checks if the Safe's threshold == 1. It should always be > 1
+            // Note - if multiple signers have already signed, then the Safe dapp doesn't do this override. So it may still fail if
+            //        number of signers == 1 < x < dynamicThresholdRequirement
+            if (safeThreshold == 1) return;
+
+            // Perhaps in future there could be some custom decoding/approvals based on the arguments too
+            threshold = getThreshold(to, bytes4(data));
```
princetonbishop, line 194: if `data.length == 0`, defaultThreshold is set anyways. Are there executions for this case? Meaning a call to a contract/function-sig combination not allowed (while `data.length == 0`. Could also be a delegate call.
Might want to safe guard against such cases.

protocol/contracts/v2/strategies/GnosisStrategy.sol
frontier159: Good spot, thanks

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
frontier159: I'm pretty sure that would be a redundant check. 
* The underlying safe threshold would be >= 1
* If the safe threshold is >= this default threshold then the guard will return regardless. 

So set to 0 or 1 has the same behaviour. Basically means by default things don't need extra signers, but you can explicity set it higher per function

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
frontier159: No i'd rather fail the whole thing (it would likely be a tenderly sim which runs first for advance notification) so they correct the issue

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
frontier159: The only thing I can think of where calldata would be null would be sending raw ETH? It's a good shout to test that explicitly too. 

Are you suggesting we should block delegatecall's? I think the safe website warns if that's the case, would need to make sure that the 'transaction batch' app still works in gnosis though as i think it uses delegatecall via the multicall contract

protocol/contracts/v2/strategies/RamosStrategy.sol
frontier159: Fixed in https://github.com/TempleDAO/temple/pull/814/commits/d5b1bc654ab8a87badb80eb957290150f4ba2c8b

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
frontier159: Not required anymore since the cooldown was rugged

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
frontier159: Not required anymore since the cooldown was rugged

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
frontier159: Fixed in https://github.com/TempleDAO/temple/pull/814/commits/2cf3760a6d25816f1b2ea81911375af2c6f4914b

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
frontier159: Fixed in https://github.com/TempleDAO/temple/pull/814/commits/2cf3760a6d25816f1b2ea81911375af2c6f4914b

protocol/contracts/v2/strategies/GnosisStrategy.sol
frontier159: Fixed in https://github.com/TempleDAO/temple/pull/814/commits/2cf3760a6d25816f1b2ea81911375af2c6f4914b

protocol/contracts/v2/TempleDebtToken.sol
```diff
@@ -0,0 +1,583 @@
...
+// Temple (v2/TempleBaseDebtToken.sol)
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { mulDiv } from "@prb/math/src/Common.sol";
+
+import { ITempleDebtToken } from "contracts/interfaces/v2/ITempleDebtToken.sol";
+import { CommonEventsAndErrors } from "contracts/common/CommonEventsAndErrors.sol";
+import { CompoundedInterest } from "contracts/v2/interestRate/CompoundedInterest.sol";
+import { CompoundedInterest } from "contracts/v2/interestRate/CompoundedInterest.sol";
```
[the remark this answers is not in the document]
frontier159: Fixed in https://github.com/TempleDAO/temple/pull/814/commits/d5b1bc654ab8a87badb80eb957290150f4ba2c8b

protocol/contracts/v2/TempleDebtToken.sol
```diff
@@ -0,0 +1,583 @@
...
+    ) {
+        uint256 _basePrincipalAndInterest = _compoundedBaseInterest();
+        return (
+            totalPrincipal,
+            _basePrincipalAndInterest - totalPrincipal,
+            estimatedTotalRiskPremiumInterest
+        );
+    }
+
+    /**
```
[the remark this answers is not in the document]
frontier159: Hmm actually - i'm not sure this will be that useful thinking about it.

The function would have to take the debtor to checkpoint. It can update that debtor's custom interest, but all the others will still be out of date.

So it still wouldn't give an accurate number. *all* users would need to be updated.

protocol/contracts/v2/TempleDebtToken.sol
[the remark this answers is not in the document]
frontier159: https://github.com/TempleDAO/temple/pull/814/commits/d57f35e43c50d4a2ae92fe07bba0e16033268f4c

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
frontier159: I've added a test and updated such that it can work for an ETH transfer - ie empty bytes payload
https://github.com/TempleDAO/temple/commit/bb87aa50d75cc0496625cba3c682a15bbf2005bb

lmk your thoughts on `delegatecall`

protocol/test/amo/ramos.ts
```diff
@@ -708,81 +901,182 @@ describe("Temple Price Floor AMO", async () => {
...
+            const totalBptStaked = await bbaUsdTempleAuraRewardPool.balanceOf(amoStaking.address);
+            await amo.setMaxRebalanceAmounts(totalBptStaked, bptIn, bptIn);
+
+            const templeTotalSupplyBefore = await templeToken.totalSupply();
+
+            await expect(amo.rebalanceUpExit(bptIn, minTempleAmountOut))
+                .to.emit(amo, "RebalanceUpExit");
+
+            // Balancer gives us a bit more Temple in the exit than what we actually asked for.
+            const actualTempleAmountOut = ethers.utils.parseEther("10.000062858762470863");
```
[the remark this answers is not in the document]
frontier159: yeh maybe, but in the following `review-feedback` pr we also check within the event too so would then need a predicate, etc. Think it's ok as is

protocol/contracts/v2/TempleDebtToken.sol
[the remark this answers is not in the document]
princetonbishop: alright, offline sounds good then

protocol/contracts/v2/safeGuards/ThresholdSafeGuard.sol
princetonbishop: commit looks good. 
oh ok, then in that case it's okay to keep support for `delegateCall`

princetonbishop (approved):

PR #798, opened 2023-04-27
Add OHM to swap options

nichosystem:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


apps/dapp/src/constants/env/preview.tsx
```diff
@@ -77,6 +78,11 @@ const env: Environment = {
       decimals: 18,
       symbol: 'OGTEMPLE',
     },
+    ohm: {
+      name: 'Olympus',
+      address: '0x64aa3364f17a4d01c6f1751fd97c2bd3d7e7f1d5',
```
marshall2112, line 83: It might be better to just leave it as an empty string, so we don't get into those issues again where we can't get balances with a bad goerli address.

(This applies to local too)

marshall2112 (changes requested):

apps/dapp/src/constants/env/preview.tsx
nichosystem: good call, [d9c28b6](https://github.com/TempleDAO/temple/pull/798/commits/d9c28b6826c6949148bd7118f0cf59a0b04165c4)

marshall2112 (approved):

PR #805, opened 2023-05-18
Add the RAMOS strategy

cool-eth:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist

### DIFF
--- a/protocol/contracts/amo/Ramos.sol
+++ b/protocol/contracts/amo/Ramos.sol
@@ -10,8 +10,6 @@ import "./interfaces/AMO__ITempleERC20Token.sol";
 import "./helpers/AMOCommon.sol";
 import "./interfaces/AMO__IAuraStaking.sol";
 
-import "forge-std/console.sol";
-
 /**
  * @title AMO built for 50TEMPLE-50BB-A-USD balancer pool
  *
@@ -319,11 +317,9 @@ contract RAMOS is Ownable, Pausable {
      * BPT tokens are then deposited and staked in Aura.
      * @param request Request data for joining balancer pool. Assumes userdata of request is
      * encoded with EXACT_TOKENS_IN_FOR_BPT_OUT type
-     * @param minBptOut Minimum amount of BPT tokens expected to receive
      */
     function addLiquidity(
-        AMO__IBalancerVault.JoinPoolRequest memory request,
-        uint256 minBptOut
+        AMO__IBalancerVault.JoinPoolRequest memory request
     ) external onlyOperatorOrOwner {
         // validate request
         if (request.assets.length != request.maxAmountsIn.length || 
@@ -337,7 +333,11 @@ contract RAMOS is Ownable, Pausable {
         AMO__ITempleERC20Token(address(temple)).mint(address(this), templeAmount);
         // safe allowance stable and TEMPLE
         temple.safeIncreaseAllowance(address(balancerVault), templeAmount);
-        stable.safeIncreaseAllowance(address(balancerVault), stableAmount);
+        if (stable.allowance(address(this), address(balancerVault)) < stableAmount) {
+            // some tokens like bb-a-USD always set the max allowance for `balancerVault`
+            // in this case, `safeIncreaseAllowance` will fail due to the arithmetic operation overflow issue
+            stable.safeIncreaseAllowance(address(balancerVault), stableAmount);
+        }
 
         // join pool
         uint256 bptAmountBefore = bptToken.balanceOf(address(this));
@@ -346,9 +346,6 @@ contract RAMOS is Ownable, Pausable {
         uint256 bptIn;
         unchecked {
             bptIn = bptAmountAfter - bptAmountBefore;
-        }
-        if (bptIn < minBptOut) {
-            revert AMOCommon.InsufficientAmountOutPostcall(minBptOut, bptIn);
         }
 
         // stake BPT
--- a/protocol/contracts/amo/Ramos.sol
+++ b/protocol/contracts/amo/Ramos.sol
@@ -267,7 +267,7 @@ contract RAMOS is Ownable, Pausable {
     function depositStable(
         uint256 amountIn,
         uint256 minBptOut
-    ) external onlyOwner whenNotPaused {
+    ) external onlyOperatorOrOwner whenNotPaused {
         _validateParams(minBptOut, amountIn, maxRebalanceAmounts.stable);
 
         stable.safeTransfer(address(poolHelper), amountIn);
@@ -296,7 +296,7 @@ contract RAMOS is Ownable, Pausable {
     function withdrawStable(
         uint256 bptAmountIn,
         uint256 minAmountOut
-    ) external onlyOwner whenNotPaused {
+    ) external onlyOperatorOrOwner whenNotPaused {
         _validateParams(minAmountOut, bptAmountIn, maxRebalanceAmounts.bpt);
 
         amoStaking.withdrawAndUnwrap(bptAmountIn, false, address(poolHelper));
@@ -322,7 +322,7 @@ contract RAMOS is Ownable, Pausable {
     function addLiquidity(
         AMO__IBalancerVault.JoinPoolRequest memory request,
         uint256 minBptOut
-    ) external onlyOwner {
+    ) external onlyOperatorOrOwner {
         // validate request
         if (request.assets.length != request.maxAmountsIn.length || 
             request.assets.length != 2 || 
@@ -362,7 +362,7 @@ contract RAMOS is Ownable, Pausable {
     function removeLiquidity(
         AMO__IBalancerVault.ExitPoolRequest memory request,
         uint256 bptIn
-    ) external onlyOwner {
+    ) external onlyOperatorOrOwner {
         // validate request
         if (request.assets.length != request.minAmountsOut.length || 
             request.assets.length != 2 || 
@@ -390,6 +390,9 @@ contract RAMOS is Ownable, Pausable {
                 unchecked {
                     receivedAmount = stable.balanceOf(address(this)) - stableAmountBefore;
                 }
+                if (receivedAmount > 0) {
+                    stable.safeTransfer(msg.sender, receivedAmount);
+                }
             }
         }
     }
@@ -402,7 +405,7 @@ contract RAMOS is Ownable, Pausable {
     function depositAndStakeBptTokens(
         uint256 amount,
         bool useContractBalance
-    ) external onlyOwner {
+    ) external onlyOperatorOrOwner {
         if (!useContractBalance) {
             bptToken.safeTransferFrom(msg.sender, address(this), amount);
         }
--- a/protocol/contracts/amo/Ramos.sol
+++ b/protocol/contracts/amo/Ramos.sol
@@ -77,7 +77,7 @@ contract RAMOS is Ownable, Pausable {
     event SetMaxRebalanceAmounts(uint256 bptMaxAmount, uint256 stableMaxAmount, uint256 templeMaxAmount);
     event WithdrawStable(uint256 bptAmountIn, uint256 amountOut, address to);
     event LiquidityAdded(uint256 stableAdded, uint256 templeAdded, uint256 bptReceived);
-    event LiquidityRemoved(uint256 stableReceuved, uint256 templeReceived, uint256 bptRemoved);
+    event LiquidityRemoved(uint256 stableReceived, uint256 templeReceived, uint256 bptRemoved);
     event SetRebalancePercentageBounds(uint64 belowTpf, uint64 aboveTpf);
     event SetTemplePriceFloorNumerator(uint128 numerator);
     event SetAmoStaking(address indexed amoStaking);
@@ -269,7 +269,7 @@ contract RAMOS is Ownable, Pausable {
     function depositStable(
         uint256 amountIn,
         uint256 minBptOut
-    ) external onlyOperatorOrOwner whenNotPaused {
+    ) external onlyOwner whenNotPaused {
         _validateParams(minBptOut, amountIn, maxRebalanceAmounts.stable);
 
         stable.safeTransfer(address(poolHelper), amountIn);
@@ -300,7 +300,7 @@ contract RAMOS is Ownable, Pausable {
         uint256 bptAmountIn,
         uint256 minAmountOut,
         address to
-    ) external onlyOperatorOrOwner whenNotPaused {
+    ) external onlyOwner whenNotPaused {
         _validateParams(minAmountOut, bptAmountIn, maxRebalanceAmounts.bpt);
 
         amoStaking.withdrawAndUnwrap(bptAmountIn, false, address(poolHelper));
@@ -326,7 +326,7 @@ contract RAMOS is Ownable, Pausable {
      */
     function addLiquidity(
         AMO__IBalancerVault.JoinPoolRequest memory request
-    ) external onlyOperatorOrOwner {
+    ) external onlyOwner {
         // validate request
         if (request.assets.length != request.maxAmountsIn.length || 
             request.assets.length != 2 || 
@@ -334,15 +334,17 @@ contract RAMOS is Ownable, Pausable {
                 revert AMOCommon.InvalidBalancerVaultRequest();
         }
 
-        uint256 templeAmount = request.maxAmountsIn[templeBalancerPoolIndex];
-        uint256 stableAmount = request.maxAmountsIn[1 - templeBalancerPoolIndex];
+        (uint256 templeAmount, uint256 stableAmount) = templeBalancerPoolIndex == 0
+            ? (request.maxAmountsIn[0], request.maxAmountsIn[1])
+            : (request.maxAmountsIn[1], request.maxAmountsIn[0]);
         AMO__ITempleERC20Token(address(temple)).mint(address(this), templeAmount);
         // safe allowance stable and TEMPLE
         temple.safeIncreaseAllowance(address(balancerVault), templeAmount);
-        if (stable.allowance(address(this), address(balancerVault)) < stableAmount) {
+        uint256 stableAllowance = stable.allowance(address(this), address(balancerVault));
+        if (stableAllowance < stableAmount) {
             // some tokens like bb-a-USD always set the max allowance for `balancerVault`
             // in this case, `safeIncreaseAllowance` will fail due to the arithmetic operation overflow issue
-            stable.safeDecreaseAllowance(address(balancerVault), stable.allowance(address(this), address(balancerVault)));
+            stable.safeDecreaseAllowance(address(balancerVault), stableAllowance);
             stable.safeIncreaseAllowance(address(balancerVault), stableAmount);
         }
 
@@ -374,7 +376,7 @@ contract RAMOS is Ownable, Pausable {
         AMO__IBalancerVault.ExitPoolRequest memory request,
         uint256 bptIn,
         address to
-    ) external onlyOperatorOrOwner {
+    ) external onlyOwner {
         // validate request
         if (request.assets.length != request.minAmountsOut.length || 
             request.assets.length != 2 || 
@@ -420,7 +422,7 @@ contract RAMOS is Ownable, Pausable {
     function depositAndStakeBptTokens(
         uint256 amount,
         bool useContractBalance
-    ) external onlyOperatorOrOwner {
+    ) external onlyOwner {
         if (!useContractBalance) {
             bptToken.safeTransferFrom(msg.sender, address(this), amount);
         }
--- a/protocol/contracts/amo/Ramos.sol
+++ b/protocol/contracts/amo/Ramos.sol
@@ -72,10 +72,12 @@ contract RAMOS is Ownable, Pausable {
     event SetPauseState(bool paused);
     event StableDeposited(uint256 amountIn, uint256 bptOut);
     event RebalanceUp(uint256 bptAmountIn, uint256 templeAmountOut);
-    event RebalanceDown(uint256 templeAmountIn, uint256 bptIn);
+    event RebalanceDown(uint256 templeAmountIn, uint256 bptOut);
     event SetPoolHelper(address poolHelper);
     event SetMaxRebalanceAmounts(uint256 bptMaxAmount, uint256 stableMaxAmount, uint256 templeMaxAmount);
-    event WithdrawStable(uint256 bptAmountIn, uint256 amountOut);
+    event WithdrawStable(uint256 bptAmountIn, uint256 amountOut, address to);
+    event LiquidityAdded(uint256 stableAdded, uint256 templeAdded, uint256 bptReceived);
+    event LiquidityRemoved(uint256 stableReceuved, uint256 templeReceived, uint256 bptRemoved);
     event SetRebalancePercentageBounds(uint64 belowTpf, uint64 aboveTpf);
     event SetTemplePriceFloorNumerator(uint128 numerator);
     event SetAmoStaking(address indexed amoStaking);
@@ -292,10 +294,12 @@ contract RAMOS is Ownable, Pausable {
      * BPT tokens are then sent into balancer pool for stable tokens in return.
      * @param bptAmountIn Amount of BPT tokens to deposit into balancer pool
      * @param minAmountOut Minimum amount of stable tokens expected to receive
+     * @param to Address to which the stable tokens withdrawn are transferred
      */
     function withdrawStable(
         uint256 bptAmountIn,
-        uint256 minAmountOut
+        uint256 minAmountOut,
+        address to
     ) external onlyOperatorOrOwner whenNotPaused {
         _validateParams(minAmountOut, bptAmountIn, maxRebalanceAmounts.bpt);
 
@@ -308,7 +312,9 @@ contract RAMOS is Ownable, Pausable {
         );
 
         lastRebalanceTimeSecs = uint64(block.timestamp);
-        emit WithdrawStable(bptAmountIn, amountOut);
+        stable.safeTransfer(to, stable.balanceOf(address(this)));
+
+        emit WithdrawStable(bptAmountIn, amountOut, to);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,137 @@
...
+     * @dev The asset value may be stale at any point in time, depending onthe strategy. 
+     * It may optionally implement `checkpointAssetBalances()` in order to update those balances.
+     */
+    function latestAssetBalances() public override view returns (AssetBalance[] memory assetBalances, uint256 debt) {
+        // RAMOS strategy assets = RAMOS's DAI balance + claimed AURA & BPT rewards =
+        // (bpt.balanceOf(RAMOS) / bpt.totalSupply() * Total_DAI_Balance in the LP) + 
+        // claimed AURA & BAL rewards
+
+        // get the RAMOS's DAI balance
+        AMO__IAuraStaking amoStaking = AMO__IAuraStaking(ramos.amoStaking());
```
frontier159, line 79: We may want to consider storing these locally as immutables to save paying for gas each time they're needed. They'll be needed for on-chain tasks too later

Downside is that if the contracts change we'll need to change here too. But I'm coming around the opinion that we could make these immutable, and if they need updating then it would mean we just push a new ramos strategy out.

Thoughts?

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,137 @@
...
+            if (asset == address(0)) {
+                // 0x00 == ETH
+                ramosStrategyBalance = 0;
+            } else if (asset == stable) {
+                // Sum the RAMOS balance and this contract's balance of the ERC20.
+                ramosStrategyBalance = stableBalanceInRamos + IERC20(asset).balanceOf(address(this));
+            } else {
+                // Sum the `amoStaking` balance, `rewardsRecipient` balance and this contract's balance of the ERC20.
+                ramosStrategyBalance = IERC20(asset).balanceOf(address(amoStaking)) + IERC20(asset).balanceOf(rewardsRecipient) + IERC20(asset).balanceOf(address(this));
+            }
```
frontier159, line 112: The rewards recipient may be a hot wallet with a bunch of different things going on. I don't think we want to include that balance. 

The assumption is that if rewards are claimed to the hot wallet, it'll be (relatively) quickly swapped to stables and that will be used to pay down the debt. So for example Lux will swap to DAI then repay the debt to TRV on behalf of the strategy.

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,137 @@
...
+        assetBalances = new AssetBalance[](length);
+
+        address asset;
+        uint256 ramosStrategyBalance;
+        for (uint256 i = 0; i < length;) {
+            asset = assets[i];
+
+            if (asset == address(0)) {
+                // 0x00 == ETH
+                ramosStrategyBalance = 0;
```
frontier159, line 105: Is the intent to ignore ETH? I think that's fine.

But if that's the case perhaps better to guard against that in the `setAssets()` function (once) rather than every time here?

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,137 @@
...
+            stableBalanceInRamos = stableBalanceInLP * bptBalanceInRamos / bptTotalSupply;
+        }
+
+        // get the RAMOS strategy assets
+        uint256 length = assets.length;
+        assetBalances = new AssetBalance[](length);
+
+        address asset;
+        uint256 ramosStrategyBalance;
+        for (uint256 i = 0; i < length;) {
```
frontier159, line 100: The assingment to 0 is unecessary since the default is already 0.
    `for (uint256 i; i < length; ++i) {...}`

(imo) the gas saved below when using `unchecked {++i}` is also very minor, but it's a negative detrement of readability. So personally don't think it's required.

protocol/contracts/v2/strategies/RamosStrategy.sol
cool-eth: Since RAMOS contract has the `setAmoStaking()` function, I think it's better to avoid the contract address mismatched by pulling `amoStaking` contract address from RAMOS contract.

protocol/contracts/v2/strategies/RamosStrategy.sol
cool-eth: Ok

protocol/contracts/v2/strategies/RamosStrategy.sol
cool-eth: Ok

protocol/contracts/v2/strategies/RamosStrategy.sol
cool-eth: Ok

protocol/contracts/v2/strategies/RamosStrategy.sol
frontier159: Sure - i'm ok either way. As an intermediate we could consider adding a single function to RAMOS which returns all addresses in one call (if needed)

Can think about it after the other functions are implemented

protocol/contracts/amo/Ramos.sol
```diff
@@ -390,6 +391,9 @@ contract RAMOS is Ownable, Pausable {
                 unchecked {
                     receivedAmount = stable.balanceOf(address(this)) - stableAmountBefore;
                 }
+                if (receivedAmount > 0) {
+                    stable.safeTransfer(msg.sender, receivedAmount);
```
frontier159, line 395: Perhaps have the receiver as a function parameter? I've found it's useful to save gas in integrations later (rather than 2 transfers)

Should we add this to withdrawStable too for symmetry?

protocol/contracts/interfaces/v2/strategies/ITempleStrategy.sol
```diff
@@ -94,12 +94,19 @@ interface ITempleStrategy is ITempleElevatedAccess {
     function currentDebt() external view returns (uint256);
 
     /**
-     * @notice How much this strategy is free to borrow from the Treasury Reserves Vault
-     * @dev This is bound by:
-     *   1/ How much stables is globally available (in the TRV + in the TRV base strategy)
-     *   2/ The amount this individual strategy is whitelisted to borrow.
-     */
-    function availableToBorrow() external view returns (uint256);
+     * @notice A strategy's current amount borrowed from the TRV, and how much remaining is free to borrow
```
frontier159, line 97: I think you need to `git pull origin temple-v2`?

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        debt = currentDebt();
+    }
+
+    /**
+     * @notice Borrow a fixed amount from the Treasury Reserves and add liquidity
+     * These stables are sent to the RAMOS to add liquidity and stake BPT
+     */
+    function borrowAndAddLiquidity(uint256 amount, uint256 slippageBps) external onlyOperator {
```
frontier159, line 149: All v2 strategies should use `onlyElevatedAccess`, please don't introduce extra roles here

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+        IERC20 bptToken = IERC20(ramos.bptToken());
+        address stable = ramos.stable();
+
+        uint256 stableBalanceInRamos;
+        {
+            uint256 bptTotalSupply = bptToken.totalSupply();
+            uint256 bptBalanceInRamos = amoStaking.stakedBalance() + bptToken.balanceOf(address(amoStaking));
+            (, uint256 stableBalanceInLP) = poolHelper.getTempleStableBalances();
+
+            stableBalanceInRamos = stableBalanceInLP * bptBalanceInRamos / bptTotalSupply;
```
frontier159, line 115: Should check if `bptTotalSupply=0`

Might be easier at that point to put this as a new view function (perhaps even within RAMOS itself?)

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * @notice Borrow a fixed amount from the Treasury Reserves and add liquidity
+     * These stables are sent to the RAMOS to add liquidity and stake BPT
+     */
+    function borrowAndAddLiquidity(uint256 amount, uint256 slippageBps) external onlyOperator {
+        // Borrow the DAI and send it to RAMOS. This will also mint `dUSD` debt.
+        treasuryReservesVault.borrow(amount, address(ramos));
+        emit Borrow(amount);
```
frontier159, line 152: This event should probably be called BorrowAndAddLiquidity, in case there are other things this strategy might need to do in the future that also borrows from TRV.

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        // Add liquidity
+        AMO__IPoolHelper poolHelper = AMO__IPoolHelper(ramos.poolHelper());
+        (,, uint256 minBptAmount, AMO__IBalancerVault.JoinPoolRequest memory requestData) = poolHelper.addLiquidityQuote(amount, slippageBps);
```
frontier159, line 156: ser I think we discussed that this shouldn't be adding directly on chain? The `requestData` should be a parameter to `borrowAndAddLiquidity()`

protocol/contracts/amo/Ramos.sol
```diff
@@ -343,9 +347,6 @@ contract RAMOS is Ownable, Pausable {
         unchecked {
             bptIn = bptAmountAfter - bptAmountBefore;
         }
-        if (bptIn < minBptOut) {
-            revert AMOCommon.InsufficientAmountOutPostcall(minBptOut, bptIn);
-        }
 
         // stake BPT
         bptToken.safeTransfer(address(amoStaking), bptIn);
```
frontier159, line 352: We might want to take the opportunity to add some new events here. eg
  `emit LiquidityAdded(stableAdded, templeAdded, bptReceived)`
Another team member (Dariox) is a subgraph king and struggled a bit with a few missing events.

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * @notice Remove liquidity and repay debt back to the Treasury Reserves
+     * The stables from the removed liquidity are sent from RAMOS to this address to repay debt
+     */
+    function removeLiquidityAndRepay(uint256 bptAmount, uint256 slippageBps) external onlyOperator {
+        // Remove liquidity
+        AMO__IPoolHelper poolHelper = AMO__IPoolHelper(ramos.poolHelper());
+        (,,,, AMO__IBalancerVault.ExitPoolRequest memory requestData) = poolHelper.removeLiquidityQuote(bptAmount, slippageBps);
```
frontier159, line 167: To also move off-chain

protocol/contracts/amo/Ramos.sol
```diff
@@ -267,7 +267,7 @@ contract RAMOS is Ownable, Pausable {
     function depositStable(
```
frontier159, line 267: ^^^ Commented here for the rebalanceDown() function.

I spotted a typo previously that we should fix while we have the opportunity, in the RebalanceDown event.

bptIn => bptOut
```
event RebalanceDown(uint256 templeAmountIn, uint256 bptOut);
```

protocol/contracts/amo/Ramos.sol
```diff
@@ -267,7 +267,7 @@ contract RAMOS is Ownable, Pausable {
...
         uint256 amountIn,
         uint256 minBptOut
-    ) external onlyOwner whenNotPaused {
+    ) external onlyOperatorOrOwner whenNotPaused {
```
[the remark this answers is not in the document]
cool-eth: Sounds good.

protocol/contracts/amo/Ramos.sol
```diff
@@ -331,9 +329,15 @@ contract RAMOS is Ownable, Pausable {
         }
 
         uint256 templeAmount = request.maxAmountsIn[templeBalancerPoolIndex];
+        uint256 stableAmount = request.maxAmountsIn[1 - templeBalancerPoolIndex];
```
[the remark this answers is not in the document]
cool-eth: Hmm... I think the current statements are more gas efficient.

protocol/contracts/amo/Ramos.sol
```diff
@@ -267,7 +267,7 @@ contract RAMOS is Ownable, Pausable {
     function depositStable(
         uint256 amountIn,
         uint256 minBptOut
-    ) external onlyOwner whenNotPaused {
+    ) external onlyOperatorOrOwner whenNotPaused {
         _validateParams(minBptOut, amountIn, maxRebalanceAmounts.stable);
 
         stable.safeTransfer(address(poolHelper), amountIn);
```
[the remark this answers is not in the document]
cool-eth: We have a test on bb-a-usd pool only. Ok. I'll fix the contract.

protocol/contracts/amo/Ramos.sol
cool-eth: Good. Will add this to `withdrawStable` too.

protocol/contracts/amo/Ramos.sol
```diff
@@ -72,10 +72,12 @@ contract RAMOS is Ownable, Pausable {
...
     event StableDeposited(uint256 amountIn, uint256 bptOut);
     event RebalanceUp(uint256 bptAmountIn, uint256 templeAmountOut);
-    event RebalanceDown(uint256 templeAmountIn, uint256 bptIn);
+    event RebalanceDown(uint256 templeAmountIn, uint256 bptOut);
     event SetPoolHelper(address poolHelper);
     event SetMaxRebalanceAmounts(uint256 bptMaxAmount, uint256 stableMaxAmount, uint256 templeMaxAmount);
-    event WithdrawStable(uint256 bptAmountIn, uint256 amountOut);
+    event WithdrawStable(uint256 bptAmountIn, uint256 amountOut, address to);
+    event LiquidityAdded(uint256 stableAdded, uint256 templeAdded, uint256 bptReceived);
+    event LiquidityRemoved(uint256 stableReceuved, uint256 templeReceived, uint256 bptRemoved);
```
frontier159, line 80: typo: `stableReceuved`

protocol/contracts/amo/Ramos.sol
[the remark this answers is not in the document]
frontier159: Were you going to update this?

protocol/contracts/amo/Ramos.sol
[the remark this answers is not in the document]
frontier159: I just checked in remix. The ternary is cheaper as I thought

```
// SPDX-License-Identifier: GPL-3.0

pragma solidity ^0.8.4;

contract Ternary {
    
    uint256 public poolIndex = 0;

    // gas==4166
    function direct(uint256[] memory theList) public view returns (uint256, uint256) {
        uint256 x = theList[poolIndex];
        uint256 y = theList[1 - poolIndex];
        return (x, y);
    }

    // gas==3880
    function ternary(uint256[] memory theList) public view returns (uint256, uint256) {
        (uint256 x, uint256 y) = poolIndex == 0 ? (theList[0], theList[1]) : (theList[1], theList[0]);
        return (x, y);
    }
}
```

protocol/contracts/amo/Ramos.sol
```diff
@@ -331,9 +335,16 @@ contract RAMOS is Ownable, Pausable {
...
 
         uint256 templeAmount = request.maxAmountsIn[templeBalancerPoolIndex];
+        uint256 stableAmount = request.maxAmountsIn[1 - templeBalancerPoolIndex];
         AMO__ITempleERC20Token(address(temple)).mint(address(this), templeAmount);
         // safe allowance stable and TEMPLE
         temple.safeIncreaseAllowance(address(balancerVault), templeAmount);
+        if (stable.allowance(address(this), address(balancerVault)) < stableAmount) {
+            // some tokens like bb-a-USD always set the max allowance for `balancerVault`
+            // in this case, `safeIncreaseAllowance` will fail due to the arithmetic operation overflow issue
+            stable.safeDecreaseAllowance(address(balancerVault), stable.allowance(address(this), address(balancerVault)));
```
frontier159, line 345: As in PoolHelper, i'm actually a bit confused by this.

protocol/contracts/interfaces/v2/strategies/ITempleStrategy.sol
frontier159: Did you do this merge?

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,186 @@
...
+     */
+    function getAssets() external view returns (address[] memory) {
+        return assets;
+    }
+
+    /**
+     * @notice Set operator
+     * @param _operator New operator
+     */
+    function setOperator(address _operator) external onlyElevatedAccess {
```
frontier159, line 81: The operator is redundant now right? Please clean up

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * @notice Remove liquidity and repay debt back to the Treasury Reserves
+     * The stables from the removed liquidity are sent from RAMOS to this address to repay debt
+     */
+    function removeLiquidityAndRepay(AMO__IBalancerVault.ExitPoolRequest memory _requestData, uint256 _bptAmount) external onlyElevatedAccess {
+        // Remove liquidity
+        ramos.removeLiquidity(_requestData, _bptAmount, address(this));
+
+        // Repay debt back to the Treasury Reserves
+        IERC20 stable = treasuryReservesVault.stableToken();
```
frontier159, line 169: There's already an immutable variable you can use, inherited from `AbstractStrategy`

`stableToken`

protocol/contracts/v2/strategies/RamosStrategy.sol
```diff
@@ -0,0 +1,186 @@
...
+
+            assetBalances[i] = AssetBalance({
+                asset: asset,
+                balance: addManualAssetBalanceDelta(ramosStrategyBalance, asset)
+            });
+        }
+
+        debt = currentDebt();
+    }
+
```
frontier159, line 146: To make the UX easier, should we add pass-through functions to the Ramos quote calls? Then the user/FE/whatever only needs to interact with the strategy contract only

frontier159 (changes requested):

protocol/contracts/amo/helpers/PoolHelper.sol
```diff
@@ -252,7 +256,10 @@ contract PoolHelper {
         AMO__IBalancerVault.JoinPoolRequest memory joinPoolRequest = createPoolJoinRequest(amountIn, joinTokenIndex, minBptOut);
 
         // approve
-        if (joinPoolToken == temple) {
+        if (joinPoolToken.allowance(address(this), address(balancerVault)) < amountIn) {
+            // some tokens like bb-a-USD always set the max allowance for `balancerVault`
+            // in this case, `safeIncreaseAllowance` will fail due to the arithmetic operation overflow issue
+            joinPoolToken.safeDecreaseAllowance(address(balancerVault), joinPoolToken.allowance(address(this), address(balancerVault)));
```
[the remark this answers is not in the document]
cool-eth: If the stable token is USDT, just increasing by the difference will not work. In this case, we need to set the allowance to 0 first and then reset it.
And I'll create a new uint256 variable for the allowance.

protocol/contracts/amo/helpers/PoolHelper.sol
```diff
@@ -335,6 +342,76 @@ contract PoolHelper {
...
+        uint256 slippageBps
+    ) external returns (
+        uint256 templeAmount,
+        uint256 expectedBptAmount,
+        uint256 minBptAmount,
+        AMO__IBalancerVault.JoinPoolRequest memory requestData
+    ) {
+        (uint256 templeBalanceInLP, uint256 stableBalanceInLP) = getTempleStableBalances();
+        templeAmount = stableBalanceInLP == 0
+            ? stablesAmount / stableWeightBps * (TPF_PRECISION - stableWeightBps)
```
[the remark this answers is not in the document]
cool-eth: Hmm... I think the weight will not affect the adding/removing liquidity and we don't need to use the `stableWeightBps`. I think we can calculate the proportional `templeAmount` using `(templeBalanceInLP * stablesAmount / stableBalanceInLP)`.
So maybe ```
templeAmount = stableBalanceInLP == 0
            ? stablesAmount
            : (templeBalanceInLP * stablesAmount / stableBalanceInLP);
```

File: protocol/contracts/amo/helpers/PoolHelper.sol (hunk 22 above)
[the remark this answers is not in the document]
Comment (frontier159) [reply protocol/contracts/amo/helpers/PoolHelper.sol:262]: Ah ok gotcha - I wasn't actually aware some tokens implemented that check. Thanks for the example

File: protocol/contracts/amo/helpers/PoolHelper.sol (hunk 23 above)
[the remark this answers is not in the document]
Comment (frontier159) [reply protocol/contracts/amo/helpers/PoolHelper.sol:359]: I *think* this is the logic which does it
https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/pool-weighted/contracts/WeightedMath.sol#L78

Called from:
https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/pool-weighted/contracts/BaseWeightedPool.sol#L111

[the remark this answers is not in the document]
Comment (frontier159) [reply protocol/contracts/amo/helpers/PoolHelper.sol:359]: Ah actually - the balancer sdk is a good resource to check too:
https://github.com/balancer/balancer-sdk/blob/be692be5d6057f5e44362667e47bd7ecf9a83b37/balancer-js/src/modules/pools/proportional-amounts/index.ts#L24

File: protocol/contracts/amo/Ramos.sol (hunk 14 above)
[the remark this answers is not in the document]
Comment (cool-eth) [reply protocol/contracts/amo/Ramos.sol:272]: Reverted the all modifier updates.

File: protocol/contracts/amo/helpers/PoolHelper.sol (hunk 23 above)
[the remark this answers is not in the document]
Comment (frontier159) [reply protocol/contracts/amo/helpers/PoolHelper.sol:359]: And just for some more context because i was poking around - the swap price calc for weighted math in typescript:
https://github.com/balancer/balancer-sor/blob/7b087e32cb9d3d24af301829b169ac88e67f1b54/src/pools/weightedPool/weightedMath.ts#L483

File: protocol/test/forge/v2/strategies/RamosStrategy.t.sol [hunk 24]
```diff
@@ -0,0 +1,458 @@
...
+        vm.expectRevert(abi.encodeWithSelector(ITempleStrategy.Unimplemented.selector));
+        strategy.automatedShutdown();
+    }
+}
+
+contract RamosStrategyTestAccess is RamosStrategyTestBase {
+    function setUp() public {
+        _setUp();
+    }
+
```
frontier159, line 96: Missing testing the access for other `onlyElevatedAccess` functions in here -- eg borrowAndAddLiquidity, removeLiquidityAndRepay

protocol/test/forge/v2/strategies/RamosStrategy.t.sol
```diff
@@ -0,0 +1,458 @@
...
+        vm.stopPrank();
+
+        // Add the strategy as an operator of RAMOS
+        address ramosOwner = RAMOS(RAMOS_ADDRESS).owner();
+        startHoax(ramosOwner);
+        RAMOS(RAMOS_ADDRESS).setOperator(address(strategy));
+        vm.stopPrank();
+    }
+
+    function getRamosBalances() internal view returns (uint256 bptTotalSupply, uint256 bptBalance, uint256 templeBalance, uint256 stableBalance) {
```
frontier159, line 332: minor - this is currently duplicated - could move into `RamosStrategyTestBase`

protocol/test/forge/v2/strategies/RamosStrategy.t.sol
```diff
@@ -0,0 +1,458 @@
...
+
+        (uint256 debt, uint256 available, uint256 ceiling) = strategy.trvBorrowPosition();
+        assertEq(debt, 0);
+        assertEq(available, borrowCeiling);
+        assertEq(ceiling, borrowCeiling);
+
+        (, uint256 bptBalanceBefore, uint256 templeBalanceBefore, uint256 stableBalanceBefore) = getRamosBalances();
+        (assetBalances,) = strategy.latestAssetBalances();
+        uint256 daiBalanceBefore = assetBalances[0].balance;
+
```
frontier159, line 359: I think it would be worth adding another `querySwap()` for a large amount of liquidity proportional to the pool size (ie 10 million worth) -- and checking that the ratio of assets before is the same as the ratio after the join (and another for exit)

It's important that we've nailed this logic or we might affect the pool in a bad way which would cost $$

protocol/test/forge/v2/strategies/RamosStrategy.t.sol
```diff
@@ -0,0 +1,458 @@
...
+        assertEq(dUSD.balanceOf(address(strategy)), amount);
+        assertEq(dUSD.balanceOf(address(trv)), 0);
+
+        (debt, available, ceiling) = strategy.trvBorrowPosition();
+        assertEq(debt, amount);
+        assertEq(available, 0.01e18);
+        assertEq(ceiling, borrowCeiling);
+
+        (, uint256 bptBalanceAfter, uint256 templeBalanceAfter, uint256 stableBalanceAfter) = getRamosBalances();
```
frontier159, line 389: I think it would be worth changing the amount added to be a large amount of liquidity proportional to the pool size (ie 10 million worth) -- or add another test for this. 5000 wei is tiny proportionally

It's important that we've nailed this logic or we might affect the pool in a bad way which would cost $$

I was contemplating doing `BalancerQueries.querySwap()`, but if we add a big amount that would give a different rate because the price impact of the next swap would be much lower.
https://docs.balancer.fi/reference/contracts/query-functions.html#queryswap

protocol/contracts/interfaces/v2/strategies/ITempleStrategy.sol
cool-eth: Had an issue on the merge. Will fix it later.

frontier159 (approved):

PR #806, opened 2023-05-18
Tlc

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,962 @@
...
+    
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                            ADMIN                           */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /**
+     * @notice Update the TLC Strategy contract, and Treasury Reserves Vault (TRV)
+     * @dev The TRV is granted access to spend DAI, in order to repay debt.
+     */
+    function setTlcStrategy(
```
cool-eth, line 397: Since `trvDebtCeiling` might be changed, maybe it's needed to call `refreshInterestRates()`

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
```diff
@@ -0,0 +1,962 @@
...
+
+        // No need to use `encodeUInt128()` here - straight from storage of the same dimension
+        _cache.interestAccumulator = uint128(debtTokenData.interestAccumulator);
+        _cache.totalDebt = debtTokenData.totalDebt;
+        _cache.interestRate = debtTokenData.interestRate;
+
+        // Set the debt ceiling and price.
+        {
+            ITreasuryReservesVault _trv = treasuryReservesVault;
+            _cache.trvDebtCeiling = _trv.strategyDebtCeiling(address(tlcStrategy));
```
cool-eth, line 669: If `trvDebtCeiling` was changed, at the end of the function, maybe it's needed to call `updateInterestRate()` and set `dirty` to `true`.

frontier159: The approach that I've taken is to expect that when the debt ceiling is increased in TRV, in order to update the interest rates, either:
1. We just wait for a new user borrow/repay/liquidation - the rate is updated at the end of those fns anyway; or
2. If we really need to 'force', the `tlc.updateInterestRates()` function can be explicitly called (can be from the same thing which updates the ceiling)

I have an existing test for (2):
https://github.com/TempleDAO/temple/blob/tlc/protocol/test/forge/v2/templeLineOfCredit/TempleLineOfCreditV2.t.sol#L473

I don't think we need to do track automatically within `initDebtTokenCache()`, because the interest rate is only used within the `initDebtTokenCache()` anyway -- and then the interest rates will be updated at the end of any borrow/repay/liquidation regardless.

Further it would make that function more complicated, and also add more storage - we'd need to need to track the debt ceiling in TLC's debt token storage too.

Let me know what you think.

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
frontier159: Done thanks

protocol/contracts/v2/templeLineOfCredit/TempleLineOfCredit.sol
cool-eth: I think both of them are fine and prefer the first option.

PR #837, opened 2023-07-26
initial import of nexus

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist

### DIFF
--- a/protocol/contracts/nexus/Relic.sol
+++ b/protocol/contracts/nexus/Relic.sol
@@ -272,9 +272,10 @@ contract Relic is
     function _beforeTokenTransfer(
         address from,
         address to,
-        uint256 tokenId
+        uint256 firstTokenId,
+        uint256 batchSize
     ) internal override(ERC721, ERC721Enumerable) whenNotPaused {
-        super._beforeTokenTransfer(from, to, tokenId);
+        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
     }
 
     function _burn(uint256 tokenId)
--- a/protocol/contracts/nexus/Relic.sol
+++ b/protocol/contracts/nexus/Relic.sol
@@ -1,0 +1,343 @@
+//                           &@@#
+//                         @@@@@@                   @@@@@@@
+//                      (@@@@@@@,                &@@@@@@@@@@@
+//                    .@@@@@@@@@                 @@@@@@@@@@@@@
+//                   @@@@@@@@@@@,                @@@@@@@@@@@@(
+//                  @@@@@@@@@@@@@                 &@@@@@@@@@
+//                 .@@@@@@@@@@@@@*
+//                 @@@@@@@@@@@@@@@/
+//                 @@@@@@@@@@@@@@@@@
+//                 @@@@@@@@@@@@@@@@@@@
+//                  @@@@@@@@@@@@@@@@@@@@
+//                  @@@@@@@@@@@@@@@@@@@@@@@*
+//                   &@@@@@@@@@@@@@@@@@@@@@@@@@@@,           .&#
+//                     @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+//                      ,@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+//                         @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+//                            #@@@@@@@@@@@@@@@@@@@@@@@*
+//                                  (@@@@@@@@@@@(
+
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import "@openzeppelin/contracts/token/ERC721/extensions/ERC721URIStorage.sol";
+import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
+import "@openzeppelin/contracts/security/Pausable.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Burnable.sol";
+import "@openzeppelin/contracts/utils/Counters.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
+
+interface IShards {
+    function equipShard(
+        address _ownerAdress,
+        uint256[] memory _itemIds,
+        uint256[] memory _amounts
+    ) external;
+
+    function unEquipShard(
+        address _target,
+        uint256[] memory _itemIds,
+        uint256[] memory _amounts
+    ) external;
+
+    function burnFromRelic(uint256 _itemId, uint256 _amount) external;
+
+    function mintFromRelic(uint256 _itemId, uint256 _amount) external;
+}
+
+contract Relic is
+    ERC721,
+    ERC721Enumerable,
+    ERC721URIStorage,
+    Pausable,
+    Ownable,
+    ERC721Burnable,
+    IERC1155Receiver,
+    ReentrancyGuard
+{
+    constructor() ERC721("Relic", "RELIC") {}
+
+    using Strings for uint256;
+    using Counters for Counters.Counter;
+    Counters.Counter public _tokenIdCounter;
+
+    enum Enclave {
+        Chaos,
+        Mystery,
+        Logic,
+        Order,
+        Structure
+    }
+    mapping(uint256 => Enclave) enclaves;
+    enum Rarity {
+        Common,
+        Uncommon,
+        Rare,
+        Epic,
+        Legendary
+    }
+    // users whitelisted for minting a Relic
+    mapping(address => bool) public whitelisted;
+    // contracts authorised to mint Relics out
+    mapping(address => bool) public whitelistedContracts;
+    // users' Relic Shard balances RELIC_ID => (SHARD_ID => BALANCE)
+    mapping(uint256 => mapping(uint256 => uint256)) public balances;
+    // relic experience points
+    mapping(uint256 => uint256) public relicXP;
+
+    // URIs served depending on relic rarity
+    mapping(Rarity => string) private BASE_URIS;
+    // shard contract interface
+    IShards public SHARDS;
+    // contract whitelisting users to mint a Relic
+    address public whitelisterAddress;
+    // XP thresholds for Relic levels
+    uint256[] public thresholds;
+
+    event MintRelic(address to, uint256 enclave);
+    event GivePoints(uint256 relicId, uint256 points);
+
+    //------- External -------//
+    // called by whitelisted users to mint Relics
+    function mintRelic(Enclave _selectedEnclave) external nonReentrant {
+        require(whitelisted[msg.sender], "You cannot own a Relic yet");
+        uint256 tokenId = _tokenIdCounter.current();
+        enclaves[tokenId] = _selectedEnclave;
+        _tokenIdCounter.increment();
+        _safeMint(msg.sender, tokenId);
+
+        whitelisted[msg.sender] = false;
+
+        emit MintRelic(msg.sender, uint256(_selectedEnclave));
+    }
+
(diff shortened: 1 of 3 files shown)


protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+
+    event Transmutation(address Templar, uint256 recipeId);
+    event MintedShard(address to, uint256 shardId, uint256 amount);
+
+    //------- External -------//
+
+    // called from Relic when transfering items from Templar wallet into Relic
+    function equipShard(address _ownerAddress, uint256[] memory _shardIds, uint256[] memory _amounts) external isRelic {
+        
+        _beforeTokenTransfer(msg.sender, _ownerAddress, address(RELIC), _shardIds, _amounts, "");
```
frontier159, line 87: This is already called within `ERC1155._safeBatchTransferFrom()`, why do you need to call it explicitly here too?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    function onERC1155Received(
+        address,
+        address,
+        uint256,
+        uint256,
+        bytes memory
+    ) public virtual override returns (bytes4) {
+        return this.onERC1155Received.selector;
```
frontier159, line 193: Should this guard to make sure the only ERC1155 it can accept is the Shards contract?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+        address,
+        address,
+        uint256[] memory,
+        uint256[] memory,
+        bytes memory
+    ) public virtual override returns (bytes4) {
+        return this.onERC1155BatchReceived.selector;
+    }
+
+    function onERC721Received(
```
frontier159, line 206: What NFTs is this Relic contract expecting to receive?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+        address,
+        address,
+        uint256,
+        bytes memory
+    ) public virtual returns (bytes4) {
+        return this.onERC721Received.selector;
+    }
+
+    function getRelicXP(uint256 _relicId) public view returns (uint256) {
```
frontier159, line 215: Fine to be here, but you already have a getter implemented for free, which would just return 0 if the relicId doesn't exist

ie this is already implemented because `relicXP` is public:

`function relicXP(uint256 _relicId) external view returns (uint256);`

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+
+    function tokenURI(uint256 _relicId)
+        public
+        view
+        virtual
+        override(ERC721, ERC721URIStorage)
+        returns (string memory)
+    {
+        require(_exists(_relicId));
+        return
```
frontier159, line 228: Can be written more simply as:
```
        return string.concat(BASE_URIS[_getRarity(_relicId)], uint256(enclaves[_relicId]).toString());
```

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    function tokenURI(uint256 _relicId)
+        public
+        view
+        virtual
+        override(ERC721, ERC721URIStorage)
+        returns (string memory)
+    {
+        require(_exists(_relicId));
```
frontier159, line 227: Missing reason string for reverting

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+            if (relicXP[_relicId] < thresholds[i]) {
+                return Rarity(i);
+            }
+        }
+        return Rarity(4);
+    }
+
+    //------- Owner -------//
+
+    function pause() public onlyOwner {
```
frontier159, line 303: Should minting/burning be allowed when paused?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+    function _beforeTokenTransfer(
+        address from,
+        address to,
+        uint256 firstTokenId,
+        uint256 batchSize
+    ) internal override(ERC721, ERC721Enumerable) whenNotPaused {
+        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
+    }
+
+    function _burn(uint256 tokenId)
```
frontier159, line 281: Nothing calls burn as far as i can tell - is there meant to be a mechanism where nexus/users can burn tokens?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+
+    function pause() public onlyOwner {
+        _pause();
+    }
+
+    function unpause() public onlyOwner {
+        _unpause();
+    }
+
+    function safeMint(address to) public onlyOwner {
```
frontier159, line 311: This seems unnecessary when you have `mintFromContract()`, and it doesn't set the enclave either?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+        uint256 tokenId = _tokenIdCounter.current();
+        _tokenIdCounter.increment();
+        _safeMint(to, tokenId);
+    }
+
+    function setThresholds(uint256[] memory _newThresholds) external onlyOwner {
```
frontier159, line 317: Good practice to emit events from these too, in case it needs to be picked up by subgraph in future

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+        thresholds = _newThresholds;
+    }
+
+    function removeFromWhitelist(address _toRemove) external onlyOwner {
```
frontier159, line 321: Access is a bit asymmetric on this - `whitelistTemplar()` is only for the whitelisterAddress. Is this intended to be onlyOwner?

protocol/contracts/nexus/Relic.sol
frontier159: Ah i missed that this implements `ERC721Burnable`. So ignore this.

Any bad consequences if a user does burn their Relic? For example shards/balances/etc would be lost too -- would this mess up balances?

protocol/contracts/nexus/TempleSacrifice.sol
```diff
@@ -0,0 +1,71 @@
...
+
+contract TempleSacrifice is Ownable, ReentrancyGuard {
+    using SafeERC20 for IERC20;
+
+    IRelic public RELIC;
+    IERC20 public TEMPLE;
+    uint256 private originTime;
+    uint256 public customPrice;
+
+    function sacrifice() external nonReentrant {
```
frontier159, line 38: What is reentrant about this function? I don't think `nonReentrant` is required?

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+    Ownable,
+    Pausable,
+    ERC1155Burnable,
+    ERC1155Supply,
+    ReentrancyGuard
+{
+    constructor() ERC1155("") {}
+
+    // whitelisted partners to mint Shards for users
+    mapping (address => bool) public partnerMinters;
```
frontier159, line 50: Do you need this `partnerMinters` mapping when you have the `partnerAllowedIds` mapping already?

If the partner is not in `partnerAllowedIds` then it won't have any ids that it's allowed to mint anyway

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+
+    function uri(uint256 _id) override public view returns(string memory){
+        return tokenURIs[_id];
+    }
+
+    //------- Owner -------//
+
+    // authorise a partner to mint 
+    function addPartner(address _partnerMinter, bool _flag) external onlyOwner {
+        partnerMinters[_partnerMinter]=_flag;
```
frontier159, line 164: Should emit events?

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+     // use receipes to transform ingredients into a new item
+    function transmute(uint256 _recipeId)
+        external
+        nonReentrant
+    {
+        require(RELIC.balanceOf(msg.sender)>0, "You must have at least one Relic to transmute Shards");
+        
+        Recipe memory transmutation = recipes[_recipeId];
+        // Destroy
+        for (uint256 i = 0; i < transmutation.requiredIds.length; i++) {
```
frontier159, line 120: I think `_burnBatch()` already checks that the sender has enough amounts?

```
            require(fromBalance >= amount, "ERC1155: burn amount exceeds balance");
```

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    function createRecipe(
+        uint256 _recipeId,
+        uint256[] memory _requiredIds,
+        uint256[] memory _requiredAmounts,
+        uint256[] memory _rewardIds,
+        uint256[] memory _rewardAmounts
+    ) external onlyOwner {
+        recipes[_recipeId].id = uint16(_recipeId);
```
frontier159, line 185: Can also write this in one go with:
```
recipes[_recipeId] = Recipe({
  requiredIds: _requiredIds,
  requiredAmounts: _requiredAmounts,
  rewardIds: _rewardIds,
  rewardAmounts: _rewardAmounts
})
```

I would also recommend adding an event here.

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+        address to,
+        uint256[] memory ids,
+        uint256[] memory amounts,
+        bytes memory data
+    ) internal override(ERC1155, ERC1155Supply) whenNotPaused {
+        super._beforeTokenTransfer(operator, from, to, ids, amounts, data);
+    }
+
+    function setURI(string memory _newUri, uint256 _index) public onlyOwner {
+        tokenURIs[_index] = _newUri;
```
frontier159, line 222: So each shard needs to have this called and set independently? After it's minted? What's the flow for calling this?

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    function pause() public onlyOwner {
```
frontier159, line 225: Should this also pause burn/minting?

protocol/contracts/nexus/Shards.sol
```diff
@@ -0,0 +1,232 @@
...
+    function unEquipShard(address _target, uint256[] memory _shardIds, uint256[] memory _amounts) external isRelic {
+        
+        _beforeTokenTransfer(address(RELIC), address(RELIC), _target, _shardIds, _amounts, "");
+
+        // transfer to target
+        _safeBatchTransferFrom( address(RELIC), _target, _shardIds, _amounts, "");
+    }
+
+    // called from Relic during Transmutations
+    function mintFromRelic(uint256 _shardId, uint256 _amount) external isRelic{
```
frontier159, line 102: mintFromRelic() and burnFromRelic() don't appear to be called from the Relic.sol contract. Should these be removed?

protocol/contracts/nexus/Relic.sol
```diff
@@ -0,0 +1,344 @@
...
+        uint256 _targetRelic,
+        uint256[] memory _itemIds,
+        uint256[] memory _amounts
+    ) external nonReentrant {
+        require(
+            msg.sender == ownerOf(_targetRelic),
+            "You don't own this Relic"
+        );
+
+        SHARDS.unEquipShard(msg.sender, _itemIds, _amounts);
```
frontier159, line 144: If the Relic contract has a bunch of shards on it for many different relic ids (ie many users), what's stopping a user unequipping any shard which it didn't equip in the first place?

Are you just relying on the fact that the below balances check reverts since you can't have a negative balance? Or some other check? Do you have a test for this?

Highly recommend using the checks-effects-interactions pattern here and anywhere else you need to transfer the NFT or ERC1155. Move any external contract interaction to the very end.

PR #843, opened 2023-08-08
Replace wagmi with web3-onboard

marshall2112:
Replace wagmi with web3-onboard.

⚠️ Some places in the code relied on certain wagmi hooks for which there is no replacement in web3-onboard. It happens that all of this functionality is no longer used in the Temple dapp, so I just commented it out / coded around it until we want to do a pass and remove all of it.


.github/workflows/dapp-tests.yaml
```diff
@@ -24,5 +24,5 @@ jobs:
           path: 'apps/dapp/node_modules'
           key: ${{ runner.os }}-modules-${{ hashFiles('apps/dapp/yarn.lock') }}
       - run: yarn install
-      - run: yarn build
+      - run: NODE_OPTIONS="--max_old_space_size=4096" yarn build
```
marshall2112, line 27: I was running into issues with memory in the action, so updated this.

apps/dapp/vite.config.ts
```diff
@@ -29,7 +29,7 @@ const plugins = [
 ];
 
 const VITE_ENV = process.env.VITE_ENV;
-const shouldBuildSourceMap = VITE_ENV === 'local' || VITE_ENV === 'preview';
+const shouldBuildSourceMap = VITE_ENV === 'local';
```
marshall2112, line 32: This is also to alleviate memory issues during build (in vercel).

nichosystem (approved):

PR #842, opened 2023-08-08
Temple v2 post audit

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist
- [ ] Code follows the style guide
- [ ] I have performed a self-review of my own code
- [ ] New and existing tests pass locally
- [ ] This PR is targeting the correct branch

PR #851, opened 2023-08-25
Fix the build

marshall2112:
I updated the monorepo / workspace configuration to remove core-subgraph (basically only have the dapp now in the apps directory included in the monorepo). Why? Because a core-subgraph dependency was causing issues. I regenerated yarn.lock, but that caused a cascade:
- Some react types were causing issues. So I added `resolutions` to package.json. 
- There was also a stray postRebalanceSlippage -> postRebalanceDelta rename from https://github.com/TempleDAO/temple/pull/842
- Once these issues were resolved, I got an issue with compatibility with BigNumbers in the vite build, with the balancer-sdk that came via the updated yarn.lock. So...
- Updated vite config fixed it

⚠️ I am essentially punting on fixing the issue with the `core-subgraph` package, so that we are unblocked. We can circle back to it when we have more time.


nichosystem:

✅ Preview
https://core-39ztuixhv-templedao.vercel.app
https://pr-851-preview.templedao.link

Built with commit 18756d9a6c8ed1410a3554b68684f0d9572c8731.

PR #852, opened 2023-08-28
V2 audit recheck

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/v2/TempleDebtToken.sol
```diff
@@ -664,6 +661,9 @@ contract TempleDebtToken is ITempleDebtToken, TempleElevatedAccess {
 
         /// @dev The risk premium interest rate for this debtor
         uint96 riskPremiumRate;
+
+        /// @dev The total balance this debtor owes: principal + baseInterest + riskPremiumInterest
+        uint128 totalBalance;
```
princetonbishop, line 666: should also update `debtorCache.totalBalance` in `_burn` for completeness

frontier159: I considered it, but it wasn't being used by anything downstream so seemed unnecessary

frontier159: Happy to take your opinion on it though?

princetonbishop: I thought the same, so tagged it as nit. let's save gas then. keep it as it is.

mirionic (approved):

PR #859, opened 2023-09-20
TLC Dapp Changes

marshall2112:
Implementing changes from the figma

- Split the features back to individual pages (instead of modals)
- Update the homepage to link to the dapp, removing the modal-related pieces


marshall2112:
Will reopen a new PR on a different branch

PR #860, opened 2023-09-25
Add OTC offer contract

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/core/OtcOffer.sol
```diff
@@ -0,0 +1,168 @@
...
+    ) {
+        userSellToken = IERC20Metadata(_userSellToken);
+        userBuyToken = IERC20Metadata(_userBuyToken);
+        fundsOwner = _fundsOwner;
+
+        offerPrice = _offerPrice;
+        offerPricingToken = _offerPricingToken;
+        offerPriceUpdateThresholdBps = _offerPriceUpdateThresholdBps;
+
+        // The price is always specified in 18dp
```
mountainpath9, line 86: It might be better to have the price expressed in the decimals of the pricing token (now that we have a nominated pricing token).

Thoughts?

These details wouldn't matter too much if there was going to be a UI for updating the price. But given that it will happen in terms of etherscan or safe tx builder, we want to minimise the risk of errors.

protocol/contracts/core/OtcOffer.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Owner can update the offer price at which user swaps can occur
+    /// @dev The new price must be within the `offerPriceUpdateThresholdBps` threshold of
+    /// the existing price
+    function setOfferPrice(uint256 _offerPrice) external onlyOwner {
+        if (_offerPrice == 0) revert CommonEventsAndErrors.ExpectedNonZero();
+
+        uint256 existingOfferPrice = offerPrice;
+        uint256 absMaxDelta = existingOfferPrice * offerPriceUpdateThresholdBps / BPS_100_PCT;
+        if (_offerPrice > existingOfferPrice) {
```
mountainpath9, line 121: Why not:

```
   if (_offerPrice > existingOfferPrice + absMaxDelta || _offerPrice < existingOfferPrice - absMaxDelta) {
       revert OfferPriceDeltaTooBig(_offerPrice, maxValid);
   }
```
less variables, gas etc?

protocol/contracts/core/OtcOffer.sol
```diff
@@ -0,0 +1,168 @@
...
+            uint256 maxValid = existingOfferPrice + absMaxDelta;
+            if (_offerPrice > maxValid) revert OfferPriceDeltaTooBig(_offerPrice, maxValid);
+        } else {
+            uint256 minValid = existingOfferPrice > absMaxDelta
```
mountainpath9, line 125: If `offerPriceUpdateThresholdBps` < BPS_100_PCT, won't existing offer price always be more than absMaxDelta?

protocol/contracts/core/OtcOffer.sol
```diff
@@ -0,0 +1,168 @@
...
+            if (_offerPrice < minValid) revert OfferPriceDeltaTooBig(_offerPrice, minValid);
+        }
+
+        offerPrice = _offerPrice;
+        emit OfferPriceSet(_offerPrice);
+    }
+
+    /// @notice Owner can update the threshold for when updating the offer price
+    function setOfferPriceUpdateThresholdBps(uint256 _offerPriceUpdateThresholdBps) external onlyOwner {
+        offerPriceUpdateThresholdBps = _offerPriceUpdateThresholdBps;
```
mountainpath9, line 137: Should this have a check < BPS_100_PCT ?

protocol/contracts/core/OtcOffer.sol
```diff
@@ -0,0 +1,168 @@
...
+        buyTokenAmount = quote(sellTokenAmount);
+
+        address _fundsOwner = fundsOwner;
+        emit Swap(msg.sender, _fundsOwner, sellTokenAmount, buyTokenAmount);
+
+        userSellToken.safeTransferFrom(msg.sender, _fundsOwner, sellTokenAmount);
+        userBuyToken.safeTransferFrom(_fundsOwner, msg.sender, buyTokenAmount);
+    }
+
+    /// @notice How many `userBuyToken` you would receive given an amount of `sellTokenAmount`
```
mountainpath9, line 154: This comment needs an update?

mountainpath9 (commented):
This contract only supports one swap pair and direction. Did you consider generalizing it to support different swaps?

Tradeoff: significant extra complexity vs having to deploy for each new swap

protocol/contracts/core/OtcOffer.sol
frontier159: In my opinion given its unlikely to have a UI, it's better to have the price decimals consistent and constant

frontier159: Since it needs to be mindless for lux and pengu

protocol/contracts/core/OtcOffer.sol
frontier159: Yeh - cheaper as 2 lines

protocol/contracts/core/OtcOffer.sol
frontier159: I'm not sure about that - what if current price is $2 and you want to set to $6

That's 300%

frontier159:
> This contract only supports one swap pair and direction. Did you consider generalizing it to support different swaps?
> 
> Tradeoff: significant extra complexity vs having to deploy for each new swap

A conscious decision to be split out and separated imo. Complexity but it also just felt right to have a seperate instance. To be seen how widely used this will be

protocol/contracts/core/OtcOffer.sol
frontier159: Also need to be sure it doesn't underflow if `absMaxDelta > existingOfferPrice` in the second branch. At which point I decided to split properly

protocol/contracts/core/OtcOffer.sol
frontier159: It can be > 100%

protocol/contracts/core/OtcOffer.sol
frontier159: Don't think so?

mountainpath9: yep - nothing to chain here.

protocol/contracts/core/OtcOffer.sol
frontier159: Can do that yep

protocol/contracts/core/OtcOffer.sol
frontier159: Non 18dp is the exception rather than the rule. So my view is that a fixed 18dp seems right

mountainpath9 (approved):
LGTM

butlerji (approved):
LGTM

PR #861, opened 2023-09-25
Fostrich tlc batch liquidate bot

frost-ostrich:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,149 @@
...
+  const txReceipt = await tx.wait();
+  const txUrl = config.CHAIN.transactionUrl(txReceipt.transactionHash);
+  // Grab the events
+  const events: TempleTaskDiscordEvent[] = [];
+  for (const ev of txReceipt?.events || []) {
+    const liquidatedEv = matchAndDecodeEvent(
+      tlc,
+      tlc.filters.Liquidated(),
+      ev
+    );
```
frost-ostrich, line 82: I'm not able to match this filter, therefore discord notification looks plain:
image

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,149 @@
...
+  const url = getChainById(config.CHAIN.id).subgraphUrl;
+  const res = await getTlcUsers(url);
+  const tlcUsers = (await res()).data?.users;
+
+  const accountsToCheck = tlcUsers?.flatMap((u) => u.id);
+  // if zero users returned from subgraph, success silently
+  if (!accountsToCheck) return taskSuccessSilent();
+  
+  checkAccPosition(accountsToCheck);
+  const tx = await tlc.batchLiquidate(accountsToCheck);
```
frost-ostrich, line 70: random error, prob 1/5 tries fail with the following:
image

image

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  const submittedAt = new Date();
+
+  const url = getChainById(config.CHAIN.id).subgraphUrl;
+  const res = await getTlcUsers(url);
+  const tlcUsers = (await res()).data?.users;
+
+  const accountsToCheck = tlcUsers?.flatMap((u) => u.id);
+  // if zero users returned from subgraph, success silently
+  if (!accountsToCheck) return taskSuccessSilent();
```
frost-ostrich, line 67: Success silently (don't send to overlord ui) when not users from subgraph

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,149 @@
...
+            liquidatedEv.daiDebtWiped,
+            18,
+            4
+          )}\``,
+        ],
+      });
+    }
+  }
+  // if no liquidation happened, success silently
+  // if (events.length === 0) return taskSuccessSilent(); // TODO: uncomment once event issue has been solved
```
frost-ostrich, line 108: Planning to success silently (don't send to overlord ui) when liquidation events are not find

apps/automation-v2/src/main.ts
```diff
@@ -0,0 +1,82 @@
...
+        );
+        return taskSuccess();
+      },
+    })
+  );
+
+  runner.main();
+}
+
+async function discordNotifyTaskException(ctx: TaskContext, te: TaskException) {
```
mountainpath9, line 57: Suggest moving this to discord.ts to keep main tidy.

apps/automation-v2/src/tlc/batch-liquidate.ts
mountainpath9: Should this hard code a gas limit rather than relying on an estimation?

apps/automation-v2/src/tlc/batch-liquidate.ts
mountainpath9: I'm unsure why this is not working. Best to add some debug prints to see manually if the event is actually present in txReceipt.events, and, assuming so, why the matching code is not recognising it.

apps/automation-v2/src/main.ts
```diff
@@ -0,0 +1,82 @@
...
+  const config = isProdnet ? CONFIG_PRODNETS : CONFIG_TESTNETS;
+
+  runner.setTaskExceptionHandler(discordNotifyTaskException);
+
+  runner.addPeriodicTask({
+    id: 'tlc-batch-liquidate',
+    cronSchedule: '*/10 * * * *',
+    action: async (ctx) => {
+      const res = await batchLiquidate(ctx, config.tlcBatchLiquidate);
+      await checkLowEthBalance(
```
mountainpath9, line 32: The checkLowEthBalance code structure started feeling a bit unwieldy to me. In swapbot, I just check for a low eth balance at the end of a successful action:

https://github.com/TempleDAO/swapbot/blob/c66e4dc71483207f2f4020681c0db7761f4b7e76/automation/src/swapbot.ts#L72

Perhaps consider this.

apps/automation-v2/src/tlc/batch-liquidate.ts
mountainpath9, line 70: batch liquidate should only be called with ids that are candidates for liquidate, and should probably chunk to a sensible max number per request.

mountainpath9 (approved):

apps/automation-v2/src/config/prodnets.ts
```diff
@@ -0,0 +1,23 @@
...
+  WALLET_NAME: 'temple_automation',
+  MIN_ETH_BALANCE: parseEther('0.1'),
+});
+
+const TLC_BATCH_LIQUIDATE_CONFIG: TlcBatchLiquidateConfig = {
+  CHAIN: MAINNET,
+  WALLET_NAME: 'temple_automation',
+  TLC_ADDRESS: '0xcbc0A8d5C7352Fe3625614ea343019e6d6b89031',
+  ACC_LIQ_MAX_CHUNK_NO: 100,
+  MIN_ETH_BALANCE: parseEther('0.05'),
```
frontier159, line 17: Does this mean we check the ETH balance both on a schedule and after the liquidate? or is this something different?

apps/automation-v2/src/main.ts
```diff
@@ -0,0 +1,47 @@
...
+  runner.setTaskExceptionHandler(discordNotifyTaskException);
+
+  runner.addPeriodicTask({
+    id: 'tlc-batch-liquidate',
+    cronSchedule: '*/10 * * * *',
+    action: async (ctx) => await batchLiquidate(ctx, config.tlcBatchLiquidate),
+  });
+
+  runner.addPeriodicTask(falliblePeriodicTask({ 
+    id: 'check-eth-balance-dsr',
```
frontier159, line 32: Is this the correct ID here? dsr should be tlc?

apps/automation-v2/src/chains.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  addressUrl(address: string) {
+    return `https://arbiscan.io/address/${address}`;
+  },
+};
+
+export const MAINNET: Chain = {
+  id: 1,
+  name: 'Mainnet',
+  subgraphUrl: 'https://api.thegraph.com/subgraphs/name/medariox/v2-mainnet',
```
frontier159, line 36: I have a feeling we may have a different subgraph url for each 'thing', it won't be common for the whole chain.

ok for now, but it might be the case that this will have to move into the actual task config instead.

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+  });
+
+  return taskSuccess();
+}
+
+const getTlcUsers = async (url: string) => {
+  const resp = await subgraphRequest<GetUserResponse>(url, {
+    query: `{
+      users {
```
frontier159, line 202: Could potentially paginate in the subgraph request too, eg

```
{
  users(first: 10, skip: 10) {
    ... on TlcUser {
      id
...
```

but given it's expected to be a low number of users, it's probably ok

if doing this, should also get the `block` on the first page, so then you use that same block for the subsequent pages to ensure consistency

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,217 @@
...
+        ... on TlcUser {
+          id
+          collateral
```
frontier159, line 205: If only using id, then probably best to just query that one field?

apps/automation-v2/src/tlc/batch-liquidate.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  if (accsToLiquidate.length === 0) return taskSuccessSilent();
+
+  // chunk compLiquidityAccs to a max number of requests per batchLiquidate
+  // e.g. try to liquidate 1000 accounts at once could use too much gas and fail
+  const accListChunks = chunkify(accsToLiquidate, config.ACC_LIQ_MAX_CHUNK_NO);
+  accListChunks.forEach(async (accBatch) => {
+    await checkAccPosition(accBatch); // check accs position before liquidation
+    const tx = await tlc.batchLiquidate(accBatch, {
+      gasLimit: 1_000_000n,
```
frontier159, line 102: Have you tested that we can liquidate 100 users for 1mm gas? I'm not sure we could.

Your test script might need to setup 100 accounts and try and liquidate them all

apps/automation-v2/src/config/prodnets.ts
frost-ostrich: yep, we check eth balance every 30 min, but also we check balance in after each batch liquidate chunk in this task, took the idea from https://github.com/TempleDAO/temple/pull/861#discussion_r1337866952

mountainpath9: I don't really see the value in checking the eth balance every 30 mins.  

Nothing except the ramos automation task can spend money from the `ramos_automation` wallet (0x67ae2161449cf3c5528fea969beba3e4f3288c61).

mountainpath9: sorry - this is not ramos, but the point remains. We will deploy a wallet specifically for these automations (to avoid nonce concurrency constraints), so no need to check except after tlc liquidate operations?

frost-ostrich: I guess this is just an extra check to inform the team in case the wallet has been used somewhere else, but... if wallet is specific for this, then no need, happy to remove if you think that's better/clearer

apps/automation-v2/src/tlc/batch-liquidate.ts
frost-ostrich: Seems 1mm is enough, however I've added gasLimit as task config
```
╰─➤  yarn local-fork:mainnet-test:tlc
yarn run v1.19.0
$ (cd ../../protocol && forge script --fork-url http://127.0.0.1:8545 scripts/deploys/localhost/v2/999-localhost-tlc-mainnet-test.s.sol)
[⠢] Compiling...
[⠒] Compiling 1 files with 0.8.19
[⠑] Solc 0.8.19 finished in 1.32s
Compiler run successful!
Script ran successfully.

== Logs ==
  No of accounts to be liquidated 1
        Running tlc.batchLiquidate...
        -total gas used           144227
        -totalCollateralClaimed   1244800000000000000
        -totalDaiDebtWiped        1065551226618879322
  
  *Move forward 1 day to trigger liquidations*

  No of accounts to be liquidated 100
        Running tlc.batchLiquidate...
        -total gas used           681405
        -totalCollateralClaimed   500000000000000000000000
        -totalDaiDebtWiped        428069788581088020488000
✨  Done in 6.38s.
```

frontier159 (approved):

PR #880, opened 2023-10-26
Pagination updates

frost-ostrich:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist
- [ ] Code follows the style guide
- [ ] I have performed a self-review of my own code
- [ ] New and existing tests pass locally
- [ ] This PR is targeting the correct branch

PR #890, opened 2023-11-07
fix: simplify react-query usage for the transaction table

mountainpath9:
Simplifications and improvements for the use of react-query in the transaction table:

* useEffect() calls no longer required to invalidate cache
* cached values are used as you page forward and back through the table
* defining the query key and the query function in the same place improves likelyhood of correctness
* simplify by having useTxHistory work in terms of row offset and limit rather than pages. This saving reexecuting useTxHistoryAvailableRows when the rowsPerPage changes.

untested. example code only

### DIFF
--- a/apps/dapp/src/components/Pages/Core/DappPages/Dashboard/Table/TxnHistoryTable.tsx
+++ b/apps/dapp/src/components/Pages/Core/DappPages/Dashboard/Table/TxnHistoryTable.tsx
@@ -1,11 +1,11 @@
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import styled from 'styled-components';
 import { TxHistoryFilterType } from '.';
 import { DashboardType } from '../DashboardContent';
 import { format } from 'date-fns';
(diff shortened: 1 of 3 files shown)


mountainpath9:
This is following the react query pattern for pagination, as described here:

https://tanstack.com/query/v4/docs/react/guides/paginated-queries

frost-ostrich:
Thx for the  added this directly to #880, the only change I didn't take from here was `defining the query key and the query function in the same place`.

The idea behind using `use-react-query.ts` is because:
-  With `getQueryKey`, we can have a single source to define all the react query keys that we will have in the app, I've noticed in the past having query keys all over the place makes it hard whenever you want to update (invalidate query) after a successful [mutation](https://tanstack.com/query/v4/docs/react/guides/mutations)
- With `useApiQuery` wrapper we can allow all our queries to use the same conf defaults, instead of defining and repeating in each

frost-ostrich:
closing in favour of #880

PR #914, opened 2023-11-28
Limit geoblocked users to legacy withdraw page

nichosystem:
# Description
Instead of preventing geoblocked users from connecting their wallet, this PR intends to limit which pages are accessible, so that only the legacy withdraw functionality is permitted. 

# Checklist


marshall2112 (approved):
LGTM

PR #965, opened 2024-02-12
multi market otc offer

princetonbishop:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+        _pause();
+    }
+
+    /// @notice Owner can unpause so user swaps can occur
+    function unpause() external override onlyElevatedAccess {
+        _unpause();
+    }
+
+    /// @notice Swap `userSellToken` for `userBuyToken`, at the `offerPrice`
+    function swap(bytes32 marketId, uint256 sellTokenAmount) external override whenNotPaused returns (uint256 buyTokenAmount) {
```
frontier159, line 130: Even for the swap functions. It does save a tiny bit of gas if the hash is done in a view first, but does it make it harder to integrate?

Maybe we could offer two variations and then user (or bot) can choose what's best for them?
`function swap(bytes32 marketId, uint256 sellTokenAmount)`
`function swap(address sellToken, address buyToken, uint256 sellTokenAmount)`

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+        if (!_otcMarketIds.contains(marketId)) { revert InvalidMarketId(marketId); }
+        if (sellTokenAmount == 0) revert CommonEventsAndErrors.ExpectedNonZero();
+        OTCMarketInfo memory marketInfo = otcMarketInfo[marketId];
+        buyTokenAmount = quote(marketId, sellTokenAmount);
```
frontier159, line 134: Calling quote here is doing more than it has to -- eg `_otcMarketIds.contains` is being called twice. I'd move the mulDiv into a separate internal function for both to use

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+ */
+contract MultiOtcOffer is IMultiOtcOffer, Pausable, ElevatedAccess {
+    using SafeERC20 for IERC20Metadata;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    /// @notice The number of decimal places represented by `offerPrice`
+    uint8 public constant OFFER_PRICE_DECIMALS = 18;
+
+    mapping(bytes32 => OTCMarketInfo) public otcMarketInfo;
+    mapping(bytes32 => OTCMarketPriceParams) public otcMarketPriceParams;
```
frontier159, line 30: I think it would be simpler if there was just one struct which contained all the market and pricing info? Having two separate mappings makes it a bit more complicated when you need all the info anyway in a swap anyway?

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+     /**
+     * @notice The available funds for a user swap is goverend by the amount of `userBuyToken` that
+     * the `fundsOwner` has available.
+     * @dev The minimum of the `fundsOwner` balance of `userBuyToken`, and the spending 
+     * approval from `fundsOwner` to this OtcOffer contract.
+     */
+    function userBuyTokenAvailable(bytes32 _marketId) external override view returns (uint256) {
+        if (!_otcMarketIds.contains(_marketId)) { revert InvalidMarketId(_marketId); }
```
frontier159, line 160: If `otcMarketInfo` and `otcMarketPriceParams` are collapsed into a single mapping, then could put this in a common internal function which does the validation?

```
function getMarketInfo(bytes32 _marketId) private view returns (MarketInfo storage) {
    if (!_otcMarketIds.contains(_marketId)) { revert InvalidMarketId(_marketId); }
    return otcMarketInfo[_marketId];
}
```

Then each function which needs can just use:
`OTCMarketInfo storage marketInfo = getMarketInfo(_marketId);`

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+        emit Swap(msg.sender, _fundsOwner, sellTokenAmount, buyTokenAmount);
+
+        _userSellToken.safeTransferFrom(msg.sender, _fundsOwner, sellTokenAmount);
+        _userBuyToken.safeTransferFrom(_fundsOwner, msg.sender, buyTokenAmount);
+    }
+
+    /// @notice How many `userBuyToken` you would receive given an amount of `sellTokenAmount`
+    function quote(bytes32 marketId, uint256 sellTokenAmount) public override view returns (uint256 buyTokenAmount) {
+        if (!_otcMarketIds.contains(marketId)) { revert InvalidMarketId(marketId); }
+        OTCMarketPriceParams memory marketPriceParam = otcMarketPriceParams[marketId];
```
frontier159, line 147: should be `storage` imo

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+        address _fundsOwner = marketInfo.fundsOwner;
+        uint256 _balance = marketInfo.userBuyToken.balanceOf(_fundsOwner);
+        uint256 _allowance = marketInfo.userBuyToken.allowance(_fundsOwner, address(this));
+        return _balance < _allowance
+            ? _balance
+            : _allowance;
+    }
+
+    function getOtcMarketIds() external view override returns (bytes32[] memory) {
+        return _otcMarketIds.values();
```
frontier159, line 171: If we had a marketId and wanted to know the tokens, is that possible? Maybe store a reverse mapping? Or is the idea that you can only know from looking at past Events?

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,187 @@
...
+        uint256 scaleDecimals = otcMarketParam.offerPricingToken == OfferPricingToken.UserBuyToken
+            ? OFFER_PRICE_DECIMALS + _otcMarketInfo.userSellToken.decimals() - _otcMarketInfo.userBuyToken.decimals()
+            : OFFER_PRICE_DECIMALS + _otcMarketInfo.userBuyToken.decimals() - _otcMarketInfo.userSellToken.decimals();
+        otcMarketParam.scalar = 10 ** scaleDecimals;
+
+        emit OtcMarketAdded(newHash, address(_otcMarketInfo.userBuyToken), address(_otcMarketInfo.userSellToken));
+    
+    }
+
+    function removeOtcMarket(bytes32 _marketId) external override onlyElevatedAccess {
```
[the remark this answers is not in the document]
princetonbishop: Okay, wil update

protocol/contracts/core/MultiOtcOffer.sol
princetonbishop: can you explain more? Because we're not changing state in this view function, just reading.

protocol/contracts/core/MultiOtcOffer.sol
princetonbishop: I was considering the setters in this decision. 2 setters `setOfferPriceRange` and `setOfferPrice` access the price params and the other access just the info. 
But I think swap is used more often so will merge the struct

protocol/contracts/core/MultiOtcOffer.sol
princetonbishop: There will be double checks on `_otcMarketIds.contains()` because if functions are changed to use `userBuyToken` & `userSellToken`,  I do validate first before getting the `marketId`

protocol/contracts/core/MultiOtcOffer.sol
princetonbishop: okay, will update

protocol/contracts/core/MultiOtcOffer.sol
frontier159: If the memory keyword is used then it does a full copy of the struct from storage into memory upfront. If you use storage keyword then it just acts like a pointer, it doesn't have to do the full copy, just access the bit it needs.

Best to compare in tests though

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,312 @@
...
+            { revert InvalidTokenPair(address(_otcMarketInfo.userBuyToken), address(_otcMarketInfo.userSellToken)); }
+        if (_otcMarketInfo.minValidOfferPrice == 0 || _otcMarketInfo.minValidOfferPrice > _otcMarketInfo.maxValidOfferPrice) 
+            { revert CommonEventsAndErrors.InvalidParam(); }
+        if (_otcMarketInfo.offerPrice < _otcMarketInfo.minValidOfferPrice || _otcMarketInfo.offerPrice > _otcMarketInfo.maxValidOfferPrice) 
+            { revert CommonEventsAndErrors.InvalidParam(); }
+
+        // check existing hash for token pair
+        marketId = _createMarketHash(address(_otcMarketInfo.userBuyToken), address(_otcMarketInfo.userSellToken));
+        if (!_otcMarketIds.add(marketId)) { revert MarketPairExists(); }
+        address[] memory tokens = new address[](2);
```
frontier159, line 62: This is only ever going to be two tokens right? Would this be better represented as a struct:
```
struct MarketTokens {
    address buyToken;
    address sellToken;
}
```

So then there's no interpretation needed by clients on what each item in the dynamic array represents?

protocol/contracts/core/MultiOtcOffer.sol
```diff
@@ -0,0 +1,312 @@
...
+        tokens[0] = address(_otcMarketInfo.userBuyToken);
+        tokens[1] = address(_otcMarketInfo.userSellToken);
+        marketIdToTokens[marketId] = tokens;
+        otcMarketInfo[marketId] = _otcMarketInfo;
+
+        OTCMarketInfo storage marketInfo = otcMarketInfo[marketId];
+        // marketInfo.minValidOfferPrice = _minValidOfferPrice;
```
frontier159, line 69: To cleanup?

protocol/contracts/core/MultiOtcOffer.sol
princetonbishop: okay will update

protocol/scripts/deploys/mainnet/v2/05-otc-offer/99-post-deploy.ts
```diff
@@ -0,0 +1,79 @@
...
+      // dai-ohm. `userBuyToken` is DAI
+      otcMarketInfo = {
+          fundsOwner: fundsOwner,
+          userBuyToken: daiToken,
+          userSellToken: ohmToken,
+          offerPricingToken: 0, // userBuyToken
+          minValidOfferPrice: ethers.utils.parseEther("11"), // minOfferPrice
+          maxValidOfferPrice: ethers.utils.parseEther("12"), //maxOfferPrice
+          scalar: 0, // default scalar
+          offerPrice: ethers.utils.parseEther("11.33") // offerPrice
```
frontier159, line 36: I think these need to be updated to be the latest values that we have onchain.

Eg for this, I think it's 11.8
https://etherscan.io/address/0x687A4B0Ac18Ed3796D55E6A1d747bD75591a8bac#readContract

Can you double check the rest too please?

frontier159 (dismissed):

frontier159 (approved):
LGTM

PR #977, opened 2024-02-29
chore: speed up deploy pipeline

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


nichosystem:

✅ Preview
https://core-6fg9s5mg8-templedao.vercel.app
https://pr-977-preview.templedao.link

Built with commit c00c51fbb5b42656354ae515972463fd6e2a4c52.

apps/dapp/vercel.json
```diff
@@ -4,5 +4,6 @@
     "silent": false,
     "enabled": false,
     "autoJobCancelation": true
-  }
+  },
+  "buildCommand": "yarn vite:build"
```
marshall2112, line 8: So now we are relying on the `prepare` step executing (more "magic") to generate the typechain artifacts. I'd personally prefer to be explicit. In other words, rename prepare to something like "gen-typechain" or whatever, and use `yarn build` as the build command (updated as `yarn gen-typechain && tsc && vite build`). Wdyt? Not a deal breaker, just curious if you agree. (On the other hand, we get the typechain for free today after yarn install, it probably improves the dev experience a bit...)

frontier159: It always was relying on `prepare` to compile and get the typechain...but it then calling this multiple times.

If you bake into `yarn build` then it's going to try and compile and copy files every singe time you want to build or `yarn dev` or whatever. The contracts won't change very frequently, so think it's good enough to:
1. Build when yarn is first setup.
2. Then call explicitly on demand if/as you need it.

marshall2112: I was thinking to only have the typechain compilation happening in certain steps, explicitly. Still, I'm fine to keep it as-is.

marshall2112 (approved):

PR #1066, opened 2024-08-08
new staking

princetonbishop:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist

### DIFF
--- a/protocol/contracts/templegold/TempleGoldStaking.sol
+++ b/protocol/contracts/templegold/TempleGoldStaking.sol
@@ -11,7 +11,6 @@ import { ITempleGoldStaking } from "contracts/interfaces/templegold/ITempleGoldS
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { Pausable } from "@openzeppelin/contracts/utils/Pausable.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import { EnumerableSet } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 
 /** 
  * @title Temple Gold Staking
@@ -22,7 +21,6 @@ import { EnumerableSet } from "@openzeppelin/contracts/utils/structs/EnumerableS
  */
 contract TempleGoldStaking is ITempleGoldStaking, TempleElevatedAccess, Pausable {
     using SafeERC20 for IERC20;
-    using EnumerableSet for EnumerableSet.UintSet;
 
     /// @notice The staking token. Temple
     IERC20 public immutable override stakingToken;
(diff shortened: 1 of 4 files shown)


protocol/contracts/templegold/TempleGoldStaking.sol
```diff
@@ -93,17 +77,15 @@ contract TempleGoldStaking is ITempleGoldStaking, TempleElevatedAccess, Pausable
...
+     * @param _period Cooldown period
      */
-    function setVestingPeriod(uint32 _period) external override onlyElevatedAccess {
-        if (_period < WEEK_LENGTH) { revert CommonEventsAndErrors.InvalidParam(); }
-        // only change after reward epoch ends
-        if (rewardData.periodFinish >= block.timestamp) { revert InvalidOperation(); }
-        vestingPeriod = _period;
-        emit VestingPeriodSet(_period);
+    function setUnstakeCooldown(uint32 _period) external override onlyElevatedAccess {
+        if (_period== 0) { revert CommonEventsAndErrors.ExpectedNonZero(); }
```
frontier159, line 86: Is this check required? If set to zero it disables the check which may be a feature?

protocol/contracts/templegold/TempleGoldStaking.sol
```diff
@@ -254,6 +232,7 @@ contract TempleGoldStaking is ITempleGoldStaking, TempleElevatedAccess, Pausable
      * @param amount Amount of staking token
      */
     function stake(uint256 amount) external override {
+        if (unstakeCooldown == 0) { revert CannotStake(); }
```
frontier159, line 235: As above - is this required? It could be a feature to disable the cooldown?

protocol/contracts/templegold/TempleGoldStaking.sol
```diff
@@ -267,20 +246,20 @@ contract TempleGoldStaking is ITempleGoldStaking, TempleElevatedAccess, Pausable
```
frontier159, line 246: I think you should remove `stakeFor()`...

The issue an auditor would find is that Bob can stakeFor() 1 wei on behalf of Alice, and that resets her clock continuously.
So she could be lcoked out for perpetuity if Bob's a real arsehole

Only other way would be for Alice to approve that Bob can stake on her behalf, first - ie she trusts Bob

protocol/contracts/templegold/TempleGoldStaking.sol
princetonbishop: agreed

protocol/contracts/templegold/TempleGoldStaking.sol
princetonbishop: Good catch

princetonbishop:
@frontier159 PR with changes [here](https://github.com/TempleDAO/temple/pull/1075)

PR #1095, opened 2024-10-10
tgld vesting contract

princetonbishop:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    /**
+     * @notice Revoke vesting
+     * @param _vestingId Vesting Id.
+     * @param _revokeTimestamp Time to revoke
+     */
+    function revokeVesting(bytes32 _vestingId, uint32 _revokeTimestamp) external override onlyElevatedAccess {
+        if (_revokeTimestamp < block.timestamp) { revert CommonEventsAndErrors.InvalidParam(); }
+        VestingSchedule storage _schedule = schedules[_vestingId];
```
frontier159, line 124: In this case, because you're accessing all the fields (and multiple times when passing into `_calculateTotalVestedAt` as `memory` anyway), it probably makes sense to use `memory` here. But will need to test the gas utilisation.

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+        _holdersVestingCount[_recipient] += 1;
+        emit RecipientChanged(_newVestingId, _oldRecipient, _recipient);
+    }
+
+    /**
+     * @notice Revoke vesting
+     * @param _vestingId Vesting Id.
+     * @param _revokeTimestamp Time to revoke
+     */
+    function revokeVesting(bytes32 _vestingId, uint32 _revokeTimestamp) external override onlyElevatedAccess {
```
frontier159, line 122: It seems like it would make things a lot simpler if you can only revoke as of 'block.timestamp'. You wouldn't need all the `revokedMaxAmounts` logic then which adds a decent amount of complexity throughout.

It would probably allow you to remove the separate `_activeVestingIds` too, since you could simply just remove from `_vestingIds`

What's the use case for revoking in a future timestamp?

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+
+    /**
+     * @notice Get vesting Id at index from list of vesting Ids
+     * @param _index Index
+     * @return id Bytes32 Id
+     */
+    function getVestingIdAtIndex(
+        uint256 _index
+    ) external view override returns (bytes32 id) {
+        if (_index >= _vestingIds.length()) { return id; }
```
frontier159, line 190: Is this check needed? It looks like it will just return zero anyway within EnumerableSet if the index isn't found?

```
    function _at(Set storage set, uint256 index) private view returns (bytes32) {
        return set._values[index];
    }
```

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+     */
+    function getTotalVestedAt(bytes32 _vestingId, uint32 _at) external view override returns (uint256) {
+        if(!_vestingIdExists(_vestingId)) { return 0; }
+        VestingSchedule memory _schedule = schedules[_vestingId];
+        if (block.timestamp <= _schedule.start) { return 0; }
+        return _calculateTotalVestedAt(_schedule, _at);
+    }
+
+    function _calculateReleasableAmount(
+        VestingSchedule memory _schedule
```
frontier159, line 391: Intended to be `memory` or `storage`? You should check other places too eg `_calculateTotalVestedAt`

protocol/contracts/interfaces/templegold/ITempleGoldVesting.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    ///@notice The owner of the TGLD funds
+    function fundsOwner() external view returns(address);
+
+    ///@notice TGLD address
+    function templeGold() external view returns(IERC20);
+
+    ///@notice Total TGLD vested and unclaimed
+    function totalVested() external view returns (uint256);
```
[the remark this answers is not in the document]
princetonbishop: good

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+            if (_schedule.cliff <= _schedule.start) { revert CommonEventsAndErrors.InvalidParam(); }
+            // total duration must be greater than cliff
+            // duration must be more than difference from start to cliff
+            if (_schedule.duration <= _schedule.cliff - _schedule.start) { revert CommonEventsAndErrors.InvalidParam(); }
+            if (_schedule.amount == 0) { revert CommonEventsAndErrors.ExpectedNonZero(); }
+           
+            bytes32 vestingId = computeNextVestingScheduleIdForHolder(_schedule.recipient);
+            schedules[vestingId] = _schedule;
+            totalVested += _schedule.amount;
+            /// @dev duplicate vesting IDs will not be generated from `computeNextVestingScheduleIdForHolder` so ignoring returned value from `add`
```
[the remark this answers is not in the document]
princetonbishop: added

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+            emit ScheduleCreated(vestingId, _schedule.recipient, _schedule.start, _schedule.cliff, _schedule.duration, _schedule.amount);
+        }
+    }
+
+    /**
+     * @notice Change recipient
+     * @param _vestingId Vesting Id.
+     * @param _recipient Recipient address
+     */
+    function changeRecipient(bytes32 _vestingId, address _recipient) external override onlyElevatedAccess {
```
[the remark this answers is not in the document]
princetonbishop: after discussions on discord, I will remove.

protocol/contracts/templegold/TempleGoldVesting.sol
princetonbishop: I use `storage` because later inthe function `_schedule.revoked=true` is written to storage.

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+        return _vestingIdExists(_id);
+    }
+
+    /**
+     * @notice Check if vesting is revoked
+     * @param _vestingId Vesting Id
+     * @return Bool
+     */
+    function isVestingRevoked(bytes32 _vestingId) external view override returns (bool) {
+        return schedules[_vestingId].revoked;
```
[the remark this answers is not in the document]
princetonbishop: you mean return true if id doesn't exist, okay sounds good

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    /**
+     * @notice Get vesting summary
+     * @param _from From timestamp
+     * @param _to To timestamp
+     * @return summary Vesting summary 
+     */
+    function getVestingSummary(uint32 _from, uint32 _to) external view override returns (VestingSummary[] memory summary) {
+        bytes32[] memory ids = _activeVestingIds.values();
```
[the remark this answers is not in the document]
princetonbishop: Valid concern. This is just for team vesting

protocol/contracts/templegold/TempleGoldVesting.sol
princetonbishop: i will make a storage version of the function. some calling functions are passing `memory`. One function passes `storage`

protocol/contracts/templegold/TempleGoldVesting.sol
princetonbishop: will move change into different branch to compare like we discussed

PR #1096, opened 2024-10-15
feat: TPI Drip initial contracts

scooprinder:
# Description
What does this PR solve?

Changes TreasuryPriceIndexOracle to instead 'drip' towards target price over a given time period. Maintains interface compatibility for `treasuryPriceIndex` but will break other usage (notably the tpiData struct).

# Checklist


protocol/contracts/interfaces/v2/ITreasuryPriceIndexOracle.sol
```diff
@@ -12,10 +12,13 @@ import { ITempleElevatedAccess } from "contracts/interfaces/v2/access/ITempleEle
  */
 interface ITreasuryPriceIndexOracle is ITempleElevatedAccess {
     event TreasuryPriceIndexSet(uint96 oldTpi, uint96 newTpi);
```
scooprinder, line 14: TODO: Remove unused events

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
```diff
@@ -46,40 +51,47 @@ contract TreasuryPriceIndexOracle is ITreasuryPriceIndexOracle, TempleElevatedAc
...
-    }
-
-    /**
-     * @notice Set the number of seconds to elapse before a new TPI will take effect.
-     */
-    function setTpiCooldown(uint32 cooldownSecs) external override onlyElevatedAccess {
-        emit TpiCooldownSet(cooldownSecs);
-        tpiData.cooldownSecs = cooldownSecs;
-    }
+        if (block.timestamp >= tpiData.targetDate) {
```
scooprinder, line 85: Pulling comment from Frontier on direct commit (which I may have nuked)

> Instead of a separate variable saying whether it's positive or negative, could tpiData.tpiSlope just be an int96 instead?
> 
> Would simplify just to this then?
> `return uint96(tpiData.currentTpi + (((tpiData.targetDate - tpiData.lastUpdatedAt) - (tpiData.targetDate - block.timestamp)) * tpiData.tpiSlope));`
> Any chance the second part is > tpiData.currentTpi (meaning this could underflow)? Or do we have guards against that elsewhere?

scooprinder, line 85: pull comment from direct commit

> Small gas optimisation:
> 
> uint32 _targetDate = tpiData.targetDate;
> And then use that local variable within. Each public variable read costs more gas than just a variable on the stack - so worth it if you access more than once in a function.

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
```diff
@@ -91,28 +103,46 @@ contract TreasuryPriceIndexOracle is ITreasuryPriceIndexOracle, TempleElevatedAc
...
+        uint96 _newTpi = targetRate;
+
+        // Check delta is within margins
         unchecked {
-            uint256 _delta = (_newTpi > _oldTpi) ? _newTpi - _oldTpi : _oldTpi - _newTpi;
-            if (_delta > maxTreasuryPriceIndexDelta) revert BreachedMaxTpiDelta(_oldTpi, _newTpi, maxTreasuryPriceIndexDelta);
+            uint256 _delta = (_newTpi > _currentTpi) ? _newTpi - _currentTpi : _currentTpi - _newTpi;
+            if (_delta > maxTreasuryPriceIndexDelta) revert BreachedMaxTpiDelta(_currentTpi, _newTpi, maxTreasuryPriceIndexDelta);
+
+            if (targetDate + uint32(block.timestamp) < maxTreasuryPriceIndexTargetDateDelta) revert BreachedMaxDateDelta(targetDate, uint32(block.timestamp), maxTreasuryPriceIndexTargetDateDelta);
```
scooprinder, line 127: Comment from direct commit

> Is this meant to be a >?

scooprinder: My understanding is that we wanted to set a minimum threshold (rather than a maximum) so `now + targetDate` must be > `maxTreasuryPriceIndexTargetDateDelta`

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
scooprinder: ACK - will adjust. Had blinkers on with this being a view but didn't consider usage in other contracts.

scooprinder: ACK - will adjust. Previously looked at int usage across the board - went with a 'simplier' approach but agree didn't hit the mark here.

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
frontier159: Should it be called `minTreasuryPriceIndexTargetDateDelta` instead then?

scooprinder: Addressed in https://github.com/TempleDAO/temple/pull/1096/commits/d1b959ae1dcd86f6b101d3b5f6f1d853880e33b5

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
scooprinder: addressed in https://github.com/TempleDAO/temple/pull/1096/commits/7effc9f07f0365a4f88fa4fcc5884d0f05ef693a

protocol/contracts/interfaces/v2/ITreasuryPriceIndexOracle.sol
scooprinder: addressed in https://github.com/TempleDAO/temple/pull/1096/commits/a68b409070b6404c75eeae9ddc8a47516d6e6d1f

scooprinder:
@frontier159 are you ok with using slither.db.json to ignore rules or would you prefer inline comments?

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
frontier159: This still seems wrong imo.

Lets say:
* targetDate = 1735732800 // 1 Jan 2025
* block.timestamp = 1729857600 // 25 Oct 2024

This is checking `1735732800 + 1729857600 < maxTreasuryPriceIndexTargetDateDelta`

Shouldn't it be:
`if (targetDate - uint32(block.timestamp) < maxTreasuryPriceIndexTargetDateDelta)`

??

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
frontier159: Done in https://github.com/TempleDAO/temple/pull/1096/commits/148c89330e66ed9987bf9c5b0d9f691e329d8cf5

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
frontier159: Fixed and tested in https://github.com/TempleDAO/temple/pull/1096/commits/148c89330e66ed9987bf9c5b0d9f691e329d8cf5

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
```diff
@@ -21,98 +20,125 @@ contract TreasuryPriceIndexOracle is ITreasuryPriceIndexOracle, TempleElevatedAc
...
-            uint256 _delta = (_newTpi > _oldTpi) ? _newTpi - _oldTpi : _oldTpi - _newTpi;
-            if (_delta > maxTreasuryPriceIndexDelta) revert BreachedMaxTpiDelta(_oldTpi, _newTpi, maxTreasuryPriceIndexDelta);
-        }
+    function setTreasuryPriceIndexAt(uint96 targetTpi, uint32 targetTime) external override onlyElevatedAccess {
+        uint96 _currentTpi = treasuryPriceIndex();
+        uint32 _now = uint32(block.timestamp);
+        int96 _tpiDelta = int96(targetTpi) - int96(_currentTpi);
+
+        // targetTime must be at or after (now + minTreasuryPriceIndexTargetTimeDelta)
+        if (targetTime < _now + minTreasuryPriceIndexTargetTimeDelta) revert BreachedMinDateDelta(targetTime, _now, minTreasuryPriceIndexTargetTimeDelta);
```
JacoboLansac, line 126: I think the `max-delta` and the `min-duration` should not be two variables to validate independently, but instead, the ratio between them is what should be validated in the requirements. The value of one without the other is meaningless, so I think it is not fully correct that their caps have individual setters (`setMaxTreasuryPriceIndexDelta()` and `setMinTreasuryPriceIndexTargetTimeDelta()`) without any constraint between them when you set them. 

Put in other words:
- If the max delta tpi is 0.2$, it means something completely different if the minTimeDelta is 1 day or 1 year.
- Similarly, if the minTimeDelta is 1 month, it means something completely different if the max delta tpi is 0.2$ or 2$

Therefore, what I would suggest is that the slope of dripping (the ratio between the tpiDelta and timeDelta) is what is validated in the requirements. 

Therefore, instead of validating the inputs against `maxTreasuryPriceIndexDelta` and `minTreasuryPriceIndexTargetTimeDelta` individually, I would have something like `max-absolute-tpi-slope`, expressed for instance like `maxTpiDeltaOverOneYear` (in absolute, as it can be negative).

Then, when you call `setTreasuryPriceIndexAt()`, your validation is not done against the absolute values of `targetTpiDelta` and `targetTimeDelta` independently, but on their RELATIONSHIP (on the slope: `targetTpiDelta` / `targetTimeDelta`). 

Example calculation below:

```solidity
	// absTpiDeltaOverOneYear = (365 days) * absTargetTpiDelta / targetTimeDelta;  // TPI slope over one year

	// the overall condition to be fulfilled in the tpi setter:
	//     require(absTpiDeltaOverOneYear < maxTpiDeltaOverOneYear);
	//     require((365 days) * absTargetTpiDelta / targetTimeDelta < maxTpiDeltaOverOneYear);
	// which can be rewritten as the following (to avoid division rounding headaches):
	require((365 days) * targetTpi < targetTimeDelta * maxTpiDeltaOverOneYear);
```

JacoboLansac: Another alternative to enforce this relationship boundaries would be to keep the two individual caps `maxTreasuryPriceIndexDelta` and `minTreasuryPriceIndexTargetTimeDelta`, and do the slope-requirement validation in both setters: 
- `setMaxTreasuryPriceIndexDelta()`
- `setMinTreasuryPriceIndexTargetTimeDelta()` 

So you know anytime you change any of the two, your know your slope is not going nuts

JacoboLansac: And of course, you can still keep the individual constrains like for min-duration if you'd like.

frontier159: Good idea, I've updated here
https://github.com/TempleDAO/temple/pull/1096/commits/27c667373ccc4302b0fe3a9f1d25acc4c5e018c3

we'd have something like:
* min 2 week duration
* max $1 TPI delta
* max slope of 1c/day

JacoboLansac: looks great

JacoboLansac (approved):
looks good to me

princetonbishop (approved):
a couple of nits. other than these lgtm

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
```diff
@@ -21,98 +20,148 @@ contract TreasuryPriceIndexOracle is ITreasuryPriceIndexOracle, TempleElevatedAc
...
+     * @notice Set the maximum absolute rate of change of TPI allowed, when 
+     * `setTreasuryPriceIndexAt()` is called.
+     * @dev Units: [TPI / second]
      */
-    function setMaxTreasuryPriceIndexDelta(uint256 maxDelta) external override onlyElevatedAccess {
-        emit MaxTreasuryPriceIndexDeltaSet(maxDelta);
-        maxTreasuryPriceIndexDelta = maxDelta;
+    function setMaxAbsTreasuryPriceIndexRateOfChange(uint96 tpiDelta, uint32 timeDelta) external override onlyElevatedAccess {
+        // Calculate the rate of change, rounding down.
+        uint96 maxAbsRateOfChange = tpiDelta / timeDelta;
```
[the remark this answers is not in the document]
frontier159: I personally see it as a bit over the top - it just means the next time `setTreasuryPriceIndexAt()` is called it will fail. This does sort of give an extra mechanism to pause any further increases which could be useful.

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
```diff
@@ -21,98 +20,148 @@ contract TreasuryPriceIndexOracle is ITreasuryPriceIndexOracle, TempleElevatedAc
...
     ) TempleElevatedAccess(_initialRescuer, _initialExecutor)
     {
         tpiData = TpiData({
-            currentTpi: _initialTreasuryPriceIndex,
-            previousTpi: _initialTreasuryPriceIndex,
-            lastUpdatedAt: uint32(block.timestamp),
-            cooldownSecs: _cooldownSecs
+            startingTpi: _initialTreasuryPriceIndex,
+            startTime: uint32(block.timestamp),
+            targetTpi: _initialTreasuryPriceIndex,
```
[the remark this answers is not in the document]
scooprinder: just to check, validate its > 0?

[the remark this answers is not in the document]
frontier159: I think it's fine to trust the deployment and post-deployment verification.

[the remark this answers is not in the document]
princetonbishop: > just to check, validate its > 0?
@scooprinder yes or if it's within range of last used value.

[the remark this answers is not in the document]
princetonbishop: > I think it's fine to trust the deployment and post-deployment verification.

okay fairs 👍

protocol/contracts/v2/TreasuryPriceIndexOracle.sol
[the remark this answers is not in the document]
princetonbishop: > I personally see it as a bit over the top - it just means the next time `setTreasuryPriceIndexAt()` is called it will fail. This does sort of give an extra mechanism to pause any further increases which could be useful.

Good point

PR #1098, opened 2024-10-21
Vesting alt

princetonbishop:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,381 @@
...
+    /// @notice TGLD address
+    IERC20 public immutable override templeGold;
+    /// @notice Total TGLD vested and unclaimed
+    uint256 public override totalVestedAndUnclaimed;
+    /// @notice Schdules for recipients
+    mapping(bytes32 id => VestingSchedule schedule) public schedules;
+    /// @notice Recipient vesting counts for generating IDs. An account can have multiple vesting schedules
+    mapping(address => uint256) private _holdersVestingCount;
+    /// @notice Maximum amount of TGLD vested for revoked accounts
+    mapping(bytes32 id => uint256 amount) public revokedMaxAmounts;
```
frontier159, line 35: Unused too - can be removed?

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,381 @@
...
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    /// @notice The owner of the TGLD funds
+    address public override fundsOwner;
+    /// @notice TGLD address
+    IERC20 public immutable override templeGold;
+    /// @notice Total TGLD vested and unclaimed
+    uint256 public override totalVestedAndUnclaimed;
+    /// @notice Schdules for recipients
+    mapping(bytes32 id => VestingSchedule schedule) public schedules;
```
frontier159, line 31: Out of interest, do you see a particular reason for using the id notation rather than a nested mapping?

eg vs:
```
mapping(address recipient => mapping(uint256 index => VestingSchedule schedule)) public schedules;
```

Just makes it a slightly less obvious API for consumers so wondering what the pros/cons are

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,381 @@
...
+        return TempleMath.mulDivRound(_schedule.amount, _elapsed, _schedule.duration, false);
+    }
+
+    /**
+     * @notice Get vesting summary
+     * @param _from From timestamp
+     * @param _to To timestamp
+     * @return summary Vesting summary 
+     */
+    function getVestingSummary(uint32 _from, uint32 _to) external view override returns (VestingSummary[] memory summary) {
```
frontier159, line 277: Would it make sense to have this function take a list of vestingId's it wants the details for?

Eg subgraph could index the `ScheduleCreated` events (and filter for the active dates), and then the client can choose if it passes in all those id's or splits it up across multiple RPC calls?

That way you can avoid any potential gas griefing issues and would make this function simpler?

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,381 @@
...
+     * @return Total vested
+     */
+    function getTotalVestedAt(bytes32 _vestingId, uint32 _at) external view override returns (uint256) {
+        VestingSchedule memory _schedule = schedules[_vestingId];
+        if (_schedule.amount == 0) { return 0; }
+        if (block.timestamp <= _schedule.start) { return 0; }
+        return _calculateTotalVestedAt(_schedule, _at);
+    }
+
+    function _calculateReleasableAmount(
```
frontier159, line 336: I don't think you need this `memory` version - the two places this is being called should be passing in `storage` too

protocol/contracts/templegold/TempleGoldVesting.sol
```diff
@@ -0,0 +1,381 @@
...
+
+    /**
+     * @notice Get vesting schedules count by recipient
+     * @param _recipient Recipient address
+     * @return Count
+     */
+    function getVestingSchedulesCountByRecipient(
+        address _recipient
+    ) external view override returns (uint256) {
+        return _holdersVestingCount[_recipient];
```
frontier159, line 146: Any reason for this vs just making `_holdersVestingCount` public?

protocol/contracts/templegold/TempleGoldVesting.sol
princetonbishop: You're right, that's more efficient

protocol/contracts/templegold/TempleGoldVesting.sol
princetonbishop: I used the id notation for the gas efficiency in read/write from storage.

PR #1105, opened 2024-10-29
feat: origami sky vault base strategy

frontier159:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist


protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+
+import { ITempleBaseStrategy } from "contracts/interfaces/v2/strategies/ITempleBaseStrategy.sol";
+import { AbstractStrategy } from "contracts/v2/strategies/AbstractStrategy.sol";
+import { CommonEventsAndErrors } from "contracts/common/CommonEventsAndErrors.sol";
+
+interface DaiUsds {
```
princetonbishop, line 13: maybe extract this into an external interface file

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    /// @notice USDS
+    IERC20 public immutable usdsToken;
+
+    /// @notice The ERC4626 vault token taking USDS deposits
+    IERC4626 public immutable usdsVaultToken;
+
+    /// @notice Maker/Sky provided contract to convert DAI<-->USDS 1:1
+    DaiUsds public immutable daiToUsds;
+
+    event DaiDeposited(uint256 amount);
```
princetonbishop, line 41: I see you don't have an interface file for this strategy yet. Perhaps a future todo?

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+        daiToUsds = DaiUsds(_daiToUsds);
+        daiToken = daiToUsds.dai();
+        usdsToken = daiToUsds.usds();
+
+        // Grant maximum approval - these can be changed later if required by elevated access
+        // see AbstractStrategy::setTokenAllowance()
+        daiToken.forceApprove(address(daiToUsds), type(uint256).max);
+        usdsToken.forceApprove(address(usdsVaultToken), type(uint256).max);
+        usdsToken.forceApprove(address(daiToUsds), type(uint256).max);
+        usdsToken.forceApprove(address(daiToUsds), type(uint256).max);
```
princetonbishop, line 62: duplicate line

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * @notice The strategy executor can shutdown this strategy, only after Governance has 
+     * marked the strategy as `isShuttingDown` in the TRV.
+     * This should handle all liquidations and send all funds back to the TRV, and will then call `TRV.shutdown()`
+     * to apply the shutdown.
+     * @dev Each strategy may require a different set of params to do the shutdown. It can abi encode/decode
+     * that data off chain, or by first calling populateShutdownData()
+     * Shutdown data isn't required for a the `usdsVaultToken` automated shutdown.
```
princetonbishop, line 215: typo `a the usdsVaultToken`

princetonbishop (commented):
left a few comments

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
frontier159: None of the strategies have interfaces as it's not expected to be used. There's a base interface for external clients to use for integration, eg [ITempleStrategy.sol](https://github.com/TempleDAO/temple/blob/stage/protocol/contracts/interfaces/v2/strategies/ITempleStrategy.sol)

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
frontier159: good spot thanks 👍

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
frontier159: Done

princetonbishop (dismissed):

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,220 @@
...
+        daiToUsds.usdsToDai(address(this), daiAmount);
+        emit DaiWithdrawn(daiAmount);
+    }
+
+    function _vaultRedemption(uint256 vaultShares) internal returns (uint256 daiAmount) {
+        // Redeem from the USDS savings vault
+        daiAmount = usdsVaultToken.redeem(vaultShares, address(this), address(this));
+
+        // Convert to DAI
+        daiToUsds.usdsToDai(address(this), daiAmount);
```
JacoboLansac, line 197: nit: 
As this function is used as part of the shutdown, you could consider converting and depositing all the DAI balance of the contract (accounting for donations). But not a big deal, since you can later call the `recoverToken()` function. So perhaps it is better to keep it clean here.

protocol/contracts/v2/strategies/SkyFarmBaseStrategy.sol
```diff
@@ -0,0 +1,220 @@
...
+     * marked the strategy as `isShuttingDown` in the TRV.
+     * This should handle all liquidations and send all funds back to the TRV, and will then call `TRV.shutdown()`
+     * to apply the shutdown.
+     * @dev Each strategy may require a different set of params to do the shutdown. It can abi encode/decode
+     * that data off chain, or by first calling populateShutdownData()
+     * Shutdown data isn't required for the `usdsVaultToken` automated shutdown.
+     */
+    function _doShutdown(bytes calldata /*data*/) internal override {
+        // Withdraw all
+        uint256 daiAmount = _vaultRedemption(usdsVaultToken.maxRedeem(address(this)));
```
JacoboLansac, line 212: It bothers me a bit that this `_vaultRedemption()` is only triggered when shutting down with `automatedShutdown()` but not if `ElevatedAccess` calls directly `TRV.shutdown()`. 

In that case, the redemption cannot be performed unless you enable the strategy again, and then do a proper shutdown with `automatedShutdown()` , so it is not the end of the world. But it bothers me so I had to point it out

JacoboLansac (approved):
solid

JacoboLansac:
I had only minor comments:
- `recoverToken()` function should restrict to make the USDS++ vault shares not recoverable.

frontier159:
> I had only a comment:
> 
> * `recoverToken()` function should restrict to make the USDS++ vault shares not recoverable.

This isn't a user facing vault/strategy - they're all protocol funds. So I don't think we need to restrict in this case. The protocol multisig can pull funds as required in any case.

mountainpath9 (approved):

PR #1139, opened 2025-01-15
cleanups and modernization of the tlc automations

mountainpath9:
goal is to structure code to fit future additions like temple gold.

Summary
 
- use the latest overlord libraries
- use the automation variable configuration
- move tlc specific code into a tlc directory


princetonbishop (approved):
new structure lgtm. I would suggest a `test/temple/temple-test.ts` file addition

PR #1170, opened 2025-03-26
Switch dapp features to chain-agnostic

marshall2112:
What does this mean? 

Basically, a similar approach that Origami takes. We use RPC providers instead of the wallet provider for reading data. This allows the dapp to function in a degraded/read-only mode even when 1) a wallet is not connected or 2) the wallet is connected to a different network/chain.

When it comes time to write a txn, we make sure a wallet is connected AND switch to the correct network.

As you can see, we have some features still using the "legacy" wrong network popover - Trade and Legacy. In the Trade page case, this is because the widget was not behaving when forcing a network switch. I am still not sure why but will continue working on this and try to find a solution. The Legacy page is rarely used and not a priority for now, and also it's more challenging to test on testnet.

Also taking this chance to remove some old features no longer used.


marshall2112:

✅ Preview
https://core-7zazwpiw4-templedao.vercel.app
https://pr-1170-preview.templedao.link

Built with commit cd23ccd64a1fc3a9440daaa653ef8a06b1cfac85.

marshall2112:
Note that all changes in this PR are also included in https://github.com/TempleDAO/temple/pull/1181 so we may ultimately decide to just merge that larger PR instead

PR #1264, opened 2026-04-21
Fix vercel deploy scripts

marshall2112:
# Description
What does this PR solve?

Fixes # (issue)

# Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated deployment pipelines for preview, staging, and production to use a unified CLI-based flow for more reliable builds and deployments.
  * Introduced deterministic preview URLs and explicit aliasing for stage/production domains.
  * Improved preview reporting by posting/updating a single PR comment with the preview URL and shortened commit reference; notifications now use the stable alias URL.

### DIFF
--- a/.github/workflows/dapp-stage.yaml
+++ b/.github/workflows/dapp-stage.yaml
@@ -22,11 +22,13 @@ jobs:
       - name: Deploy to Vercel
         id: deploy_step
         run: |
-          npx vercel@52.0.0 pull --yes --environment=preview -A apps/dapp/vercel.json
-          npx vercel@52.0.0 build -A apps/dapp/vercel.json
-          DEPLOYMENT_URL=$(npx vercel@52.0.0 deploy --prebuilt -A apps/dapp/vercel.json | tail -1)
+          set -euo pipefail
+          vcl() { npx --yes vercel@52.0.0 -A apps/dapp/vercel.json "$@"; }
+          vcl pull --yes --environment=preview
+          vcl build
+          DEPLOYMENT_URL=$(vcl deploy --prebuilt)
           echo "deployment_url=$DEPLOYMENT_URL" >> $GITHUB_OUTPUT
-          [ -n "$DEPLOYMENT_URL" ] && npx vercel@52.0.0 alias set "$DEPLOYMENT_URL" "stage.templedao.link" --scope "$VERCEL_ORG_ID" || exit 1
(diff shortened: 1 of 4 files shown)


.github/workflows/dapp-stage.yaml
```diff
@@ -6,23 +6,27 @@ on:
...
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          github-comment: false
-          scope: ${{ secrets.VERCEL_ORG_ID }}
-          alias-domains: | #Optional
-            stage.templedao.link
+          node-version: '24'
+      - name: Deploy to Vercel
+        id: deploy_step
+        run: |
+          npx vercel@52.0.0 pull --yes --environment=preview -A apps/dapp/vercel.json
```
pocin: i think this can be simplified a bit using a function to avoid repetition, something along the lines of this?
```sh
  set -euo pipefail
  vcl() {
    npx --yes vercel@52.0.0 -A apps/dapp/vercel.json "$@"
  }

  vcl pull --yes --environment=production
  vcl build --prod
  DEPLOYMENT_URL=$(vcl deploy --prebuilt --prod)
  vcl alias set "$DEPLOYMENT_URL" "templedao.link" --scope "$VERCEL_ORG_ID"
```

also here vercel build stdout always contains the DEPLOYMENT_URL so we can simplify a bit further https://vercel.com/docs/cli/deploy#standard-output-usage

```sh
  DEPLOYMENT_URL=$(vcl deploy --prebuilt --prod)
```
captures only stdout (and thanks to `set -e` on top, fails on error (with stderr in the logs)

pocin: we might also be able to cache the vercel  binary using `actions/cache@v5` but i think we can defer that if the set-up is actually slow

pocin (changes requested):

.github/workflows/dapp-stage.yaml
marshall2112: > we might also be able to cache the vercel binary using `actions/cache@v5` but i think we can defer that if the set-up is actually slow

Yes I agree, a good step but we can defer and see if it's needed.

marshall2112: https://github.com/TempleDAO/temple/pull/1264/commits/ea5acc301698f1cc766c2da5929f9920c539c99b

pocin (approved):

PR #1279, opened 2026-05-14
chore: package manager and CI hardening

pocin:
# Description
A mix of gpt guided security hardening, along with inputs from https://astral.sh/blog/open-source-security-at-astral (mention https://docs.zizmor.sh for static analysis of CI config)

```
zizmor --collect=workflows .github
```
from the root of the repo.

**Split into logical commits with descriptions, read them for more context.**

Two side notes:

1) there is a sync-typechain-dirs.yml which i dare not to touch that has write access to the repo

2) protocol-tests specify `security-events: true` perms. I dont think they are required, but kept them 
https://github.com/TempleDAO/temple/blob/3ad95d669d2d95586050ea00edafd13a34ed91ea/.github/workflows/protocol-tests.yaml#L17-L18

3) deliberately skips `setup-node` caching step for prod deployments. If speed becomes issue, i think we should first port to blacksmith, then think about enabling caches

Fixes # (issue)

# Checklist

## Summary by CodeRabbit

* **Chores**
  * CI security tightened: checkout credentials isolated and job permissions limited.
  * Deploys now use a locally-resolved Vercel CLI and explicit deploy env/config.
  * Preview workflow passes preview alias/URL via env for accurate preview links.
  * Deterministic installs: frozen lockfiles, non-interactive installs, and lifecycle scripts disabled.
  * Added explicit packageManager entries across projects.
  * CI installs app dependencies and preserves Yarn config/artifacts in build containers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TempleDAO/temple/pull/1279?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/vercel-preview.yaml
+++ b/.github/workflows/vercel-preview.yaml
@@ -14,6 +14,8 @@ jobs:
       VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
       VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
       VERCEL_PROJECT_ID: ${{ secrets.DAPP_PROJECT_ID }}
+      PREVIEW_ALIAS: pr-${{ github.event.number }}-preview.templedao.link
+      PREVIEW_URL: https://pr-${{ github.event.number }}-preview.templedao.link
     steps:
       - name: checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
@@ -27,14 +29,13 @@ jobs:
         working-directory: apps/dapp
       - name: Deploy to Vercel
         id: deploy_step
-        run: bash .github/scripts/vercel-deploy.sh preview pr-${{ github.event.number }}-preview.templedao.link
+        run: bash .github/scripts/vercel-deploy.sh preview "$PREVIEW_ALIAS"
         continue-on-error: true
       - name: Comment on PR
         if: steps.deploy_step.outcome == 'success'
         uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b
         env:
           DEPLOYMENT_URL: ${{ steps.deploy_step.outputs.deployment_url }}
-          PREVIEW_URL: https://pr-${{ github.event.number }}-preview.templedao.link
         with:
           script: |
             const marker = '<!-- vercel-preview-comment -->';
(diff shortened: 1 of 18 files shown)


.github/workflows/vercel-preview.yaml
```diff
@@ -17,21 +17,29 @@ jobs:
...
+          persist-credentials: false
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e
         with:
           node-version: '24'
+      - name: Install Dependencies
+        run: yarn install
+        working-directory: apps/dapp
       - name: Deploy to Vercel
         id: deploy_step
         run: bash .github/scripts/vercel-deploy.sh preview pr-${{ github.event.number }}-preview.templedao.link
```
marshall2112: Should we do the same here, replacing the in-line template expression with an `env` var?

apps/dapp/.npmrc
```diff
@@ -0,0 +1,4 @@
+ignore-scripts=true
+allow-git=none
+min-release-age=3
```
marshall2112: I think we'll need `engine-strict=true, along with forcing npm at least version 11.10.0 for this to work as expected.

apps/dapp/package.json
marshall2112, line 5: Do we need to specify `"npm": ">=11.10.0"` here so we get access to `min-release-age`? Afaik, this is the min version required for this to work.

Alongside this maybe also adding `engine-strict=true` in npmrc.

marshall2112 (changes requested):

.github/workflows/vercel-preview.yaml
pocin: pushed a commit which avoids repetition and references an env var

apps/dapp/.npmrc
pocin: Yeah seems like you are right, but I have actually removed the .npmrc files. 

1) The repo is yarn based, so this should have no effct in real use case.
There are still some `npx hardhat` sprinkled around in the protocol repo, but these should resolve to the local node_modules/hardhat installed as dev depencency.

It was added while we still used npx vercel and left in as a safety net just in case but i dont think it's relevant anymore

marshall2112: Yeah the repo is yarn based, I thought we wanted the .npmrc in case someone accidentally used npm instead. But then I realized we could force yarn using devEngines field? https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines

apps/dapp/package.json
marshall2112: On further reading, we could use `devEngines` I think

marshall2112 (dismissed):

marshall2112 (approved):

PR #1296, opened 2026-07-07
Dependency updates to resolve additional security alerts

marshall2112:
# Description

Resolves the last of the high dependabot security alerts.

Note that we have a Vite major upgrade here (5.4 to 6.4.3, plus @vitejs/plugin-react to 4.7.0 and @vitejs/plugin-legacy to 6.1.1). No changes to `vite.config.ts` were required as the existing config is fully valid vite 6 API. 

Verified yarn build passes end-to-end (tsc + vite), legacy chunks and `nomodule` scripts are still emitted, and the dev server starts and serves normally (plus the preview deploy also worked).


princetonbishop (approved):
approved. just something minor to check: that vite@6.4.3 itself doesn't break dapp's config

